PR #15, opened 2017-04-30
Exchange 'internal' option & exchange-to-exchange bindings

juntalis:
Recently had a requirement for this in one of my projects, so I figured I'd submit the change. Clearly just a copy-paste job on the ``Queue.bind``/``Queue.unbind`` methods. I can add some tests for the changes if you'd like - just wasn't sure whether you preferred it get its own separate test case.

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -39,6 +39,7 @@ class Channel(BaseChannel):
             ExchangeType.DIRECT,
             durable=None,
             auto_delete=None,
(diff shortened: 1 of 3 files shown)


juntalis:
Sorry about that - making these changes from memory without a local test environment setup. Will return to address that last issue when I wake up in a few hours.

mosquito:
Looks great I will start review tomorrow. Thank you.

PR #29, opened 2017-05-28
Fix channel close callback

decaz:
I think it concerns #24 as in case of closing channel there can be raised InvalidStateError on removed string.


decaz:
@mosquito any thoughts?

mosquito:
Cool. Sorry for delay.

PR #36, opened 2017-06-20
Feature/queue cancel support

### LINKED ISSUE #35: `basic_cancel` is not implemented
Hi there,

As far as I can see in the code and in the documentation, [`basic_cancel`](https://github.com/pika/pika/blob/51fd186f1fabe0da8582609d266df1e28e57e3f4/pika/channel.py#L202) is not implemented in `aio-pika`. Moreover, it's currently not possible to hack around this, since aio-pika's `consume` method doesn't return the consumer tag.

May I submit a PR for this?

Thanks,
Gilb's

mosquito:
Should close #35

### DIFF
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -173,8 +173,7 @@ class Queue(BaseChannel):
         )
 
     @BaseChannel._ensure_channel_is_open
-    @asyncio.coroutine
-    def cancel(self, consumer_tag: ConsumerTag, timeout=None):
+    def cancel(self, consumer_tag: ConsumerTag, timeout=None, nowait: bool=False):
         """ This method cancels a consumer. This does not affect already
         delivered messages, but it does mean the server will not send any more
         messages for that consumer. The client may receive an arbitrary number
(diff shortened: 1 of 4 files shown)


aio_pika/queue.py
```diff
@@ -170,6 +174,28 @@ def consumer(channel: Channel, envelope, properties, body: bytes):
...
+        for any reason other than the server receiving the corresponding
+        basic.cancel from the client). This allows clients to be notified of
+        the loss of consumers due to events such as queue deletion.
+
+        :param consumer_tag: consumer tag returned by :func:`~aio_pika.Queue.consume`
+        :param timeout: execution timeout
+        :return: Basic.CancelOk when operation completed successfully
+        """
+        f = self._create_future(timeout)
+        self._channel.basic_cancel(f.set_result, consumer_tag=consumer_tag, nowait=False)
```
gilbsgilbs: I think you set `nowait` to `false` to prevent the future from never be callbacked by pika? Don't you think this option should be settable by the user? (e.g. `if nowait: f.set_result(None)`? That might be a bit confusing though.)

mosquito: When broker not responses future will be cancelled by timeout.

mosquito: Otherwise you don't have any possibility to know a result.

mosquito: But actually you are right. The `consue` has a blocking interface and `cancel` should have the same variant.

gilbsgilbs: You're right, I don't see any use-case. Still the option is present in Pika and I don't see any reason for aio-pika to prevent the user from setting the option if he really wants to. That's why I was suggesting to `set_result` manually when `nowait` is truthy. 

```
f = self._create_future(timeout)
self._channel.basic_cancel(f.set_result, consumer_tag=consumer_tag, nowait=nowait)
if nowait:
    f.set_result(None)
return (yield from f)
```

Indeed that's not really ideal, I'm not even sure if it would work, but at least, it does not limitate use cases.

gilbsgilbs:
That's great job. Thanks @mosquito for your impressive reactivity!

mosquito:
@gilbsgilbs my pleasure

PR #38, opened 2017-06-21
[fea] add queue message_count and consumer_count

### LINKED ISSUE #37: Expose message/consumer counts from declare_queue
Due to the `Queue` abstractions it's currently impossible to safely get the [actual result](https://github.com/pika/pika/blob/0.10.0/pika/spec.py#L1014) of the `queue.declare` operation, along with the `message_count` and `consumer_count`.

A trivial solution would be to add `message_count` and `consumer_count` to `Queue`, and then set it [here](https://github.com/mosquito/aio-pika/blob/master/aio_pika/queue.py#L66) along with `name`. Would this be an acceptable approach for a PR?

connec:
Fixes #37.

### DIFF
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -18,6 +18,7 @@ class Queue(BaseChannel):
 
     __slots__ = ('name', 'durable', 'exclusive',
                  'auto_delete', 'arguments',
+                 'message_count', 'consumer_count',
                  '_channel', '__closing')
 
     def __init__(self, loop: asyncio.AbstractEventLoop, future_store: FutureStore,
@@ -31,6 +32,8 @@ class Queue(BaseChannel):
         self.exclusive = exclusive
         self.auto_delete = auto_delete
         self.arguments = arguments
+        self.message_count = None
+        self.consumer_count = None
 
     def __str__(self):
         return "%s" % self.name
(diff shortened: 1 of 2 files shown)


aio_pika/queue.py
```diff
@@ -63,7 +66,10 @@ def declare(self, timeout: int = None, passive: bool = False) -> asyncio.Future:
         )
 
         def on_queue_declared(result):
-            self.name = result.result().method.queue
+            method = result.result().method
+            self.name = method.queue
+            self.message_count = method.message_count
+            self.consumer_count = method.consumer_count
```
mosquito: I guess you should increment it after `consume` and decrement when `cancel`.

mosquito: But actually it's wrong. This is server-side state.

connec: Yes, it's server-side state that gets returned with every `queue-ok` response from the AMQP server.

It does seem a bit odd putting it on `Queue` (since it will be out-of-date as soon as it's used), but it's the easiest thing to do.

An alternative would be to expose the underlying `Queue.declare` future somehow. Obviously it would be possible to 'manually' call `Queue.declare` but then you also have to manually take care of all the [setup in Channel](https://github.com/mosquito/aio-pika/blob/master/aio_pika/channel.py#L195).

connec: An example alternative would be to add a `__slot__` called, e.g., `declare_result` and store the entire `queue-ok` object in there.

connec: Actually, given the attempt of the library to have a `Queue` abstraction I think storing them on the `Queue` makes the most sense (especially since there's no way of creating a `Queue` except by `channel.declare_queue`). They could perhaps be renamed to `last_message_count`/`last_consumer_count` or something, to make it clear that the values may be out of date.

mosquito:
Thank you for your idea. I already fix this in 0.15.0.

PR #43, opened 2017-07-08
Prevent error log for normal channel close.

### LINKED ISSUE #40: Error log for normal shutdown
Hi,

When connection is closed normally, aio_pika outputs an error log invariably : `aio_pika.channel - ERROR - Channel <pika.channel.Channel object at 0x7fdc76040c50> closed: 0`. See https://github.com/mosquito/aio-pika/blob/65de58dbbe75fee0a95ce947ef6e73ab00c3dec5/aio_pika/channel.py#L66 .

This should happen only when the connection is closed abnormally.

gilbsgilbs:
Fixes #40


mosquito:
Thank you for your contribution.

PR #48, opened 2017-07-19
Add note about asyncio event loop

maxnordlund:
This is a hard requirement and will probably get new developers if not mentioned early.

### DIFF
--- a/docs/source/rabbitmq-tutorial/1-introduction.rst
+++ b/docs/source/rabbitmq-tutorial/1-introduction.rst
@@ -75,7 +75,14 @@ Hello World!
 ++++++++++++
 
 .. note::
-    Using the `aio-pika`_ async Python client
+    Using the `aio-pika`_ async Python client. To use you need a `asyncio eventloop`_, for
+    example by appending:
+
+    .. literalinclude:: examples/1-introduction/send.py
+        :language: python
+        :lines: 22-24
+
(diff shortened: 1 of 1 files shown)


mosquito:
By default, event `loop` argument not required. If loop not passed will be called `asyncio.get_event_loop()`

maxnordlund:
But you still need the `loop.run_until_complete`, right?

mosquito:
Lets add it into examples.

maxnordlund:
I don't quite follow, do you mean add a new example or something else?

mosquito:
Ok, my point is [this example](https://aio-pika.readthedocs.io/en/latest/rabbitmq-tutorial/1-introduction.html#putting-it-all-together) already has it. 
If you want to add some info about event loop (and asyncio of course), just add it as a different note block. Add link to the [asyncio tutorial](https://pymotw.com/3/asyncio/coroutines.html) or [python documentation](https://docs.python.org/3/library/asyncio.html)

mosquito:
@maxnordlund what did you think [changes in warning block](https://aio-pika.readthedocs.io/en/latest/rabbitmq-tutorial/1-introduction.html#introduction) closed this issue?

maxnordlund:
Yeah that's better, but I do think a reminder in the quickstart doesn't hurt. Feel free to close this if you want.

PR #49, opened 2017-07-20
ability to process message yourself under message.process context

smagafurov:


mosquito:
Actually it's so useful. But I have some comments about naming and semantic:
0. When you try to process locked message, you should have an exception (it's correct in this PR)
1. `ignore_processed=False` it's more speaking name.
2. You should add some description about use cases. (e.g. "This useful when you share message instance between coroutines"). BTW: I'm not sure that I correctly understood your intentions 😀.

smagafurov:
My use case is very simple:

```python
with message.process():
    response = await json_rpc_handler(message)
    if response.data.get('error'):
        message.reject()  # move message to DLX
    ...
```
I don't want to raise exception here
I just want to reject message in some cases

smagafurov:
fixed

PR #50, opened 2017-07-20
support channels without confirmation (in pursuit of performance)

smagafurov:


mosquito:
Thank you for your contribution. But `no_confirm` is a little bit messy. Can you name it like in rabbitmq: `publisher_confirms=True` by default?

smagafurov:
fixed

PR #51, opened 2017-07-27
channel public closing future

smagafurov:


smagafurov:
@mosquito , hi!
Can't understand travis problem with py 3.4
Can you explain?
Why problem is only with py34?

smagafurov:
I got it!
fixed

PR #53, opened 2017-08-30
Bump pika version to 0.11

decaz:


mosquito:
Hi @decaz please bump the `aio-pika`'s version

decaz:
@mosquito done!

mosquito:
@decaz in the [pika's 0.11 history](https://pika.readthedocs.io/en/0.11.0/version_history.html) was added [asyncio connection-adapter](https://github.com/pika/pika/pull/805) based on my implementation.

As a complete solution we are have to remove redundant code from aio-pika.

mosquito:
@decaz I guess the simplest solution is remove [`IOLoopAdapter`](https://github.com/mosquito/aio-pika/blob/master/aio_pika/adapter.py#L15-L115) and inherit pika's connection adapter only for [set client_properties](https://github.com/mosquito/aio-pika/blob/master/aio_pika/adapter.py#L159-L173)

decaz:
@mosquito I agree, added some changes.

setup.py
```diff
@@ -10,7 +10,7 @@
 
 requires = [
     'shortuuid',
-    'pika<0.11',
+    'pika<0.12',
```
mosquito: and `>=0.11` is required now.

decaz: Fixed.

mosquito:
Excellent. Thank you for your contribution.

mosquito:
The #54 stops reproduction after revert to pika-0.10.

I have to revert this changes. I can't imagine test-case for this and I have to watch similar pika's issues.

I'm sorry.

mosquito:
Actually you can use 0.20.x releases. 😁

Issue #56, opened 2017-09-07
await queue.get() seems to gets stuck when the queue is empty (not getting new elements)


jakubczaplicki:
What should be the correct behaviour in the following case :

```
async def consumer(queue_in):
    while True:
        msg = await queue_in.get()
        await do_stuff(msg)
```

From what I observe is that the queue just gets stuck waiting for an element. But when an element is added to the queue, the `get()` doesn't "see" it, and keeps waiting: `2017-09-07 15:13:50,959 aio_pika.queue  DEBUG    Awaiting message from queue: <Queue(queue_in): auto_delete=False, durable=None, exclusive=False, arguments=None>`

However after adding the timeout:

```
async def consumer(queue_in):
    while True:
        try:
            msg = await queue_in.get(timeout=3.0)
            await do_stuff(msg)
        except TimeoutError:
            await asyncio.sleep(1.5)
```

The `get()` times out (even when an element is added to the queue within the timeout), and in the next iteration it reads all the elements from the queue added in the meantime.

Am I doing something wrong ?

mosquito:
Hello. Please try to reproduce it on `0.21.0` this might be because #54

jakubczaplicki:
I am afraid it still doesn't work with `0.21.0`

This is the script I use to reproduce the issue:  https://gist.github.com/jakubczaplicki/8993755aa70a73c506c07a05c6f65da1 (make sure to change <user>, <pswd>, etc.)

Running `python aio_pika_get_test.py consume` , the script uses queue.consume method and works as expected:
```
Using queue.consume
DEBUG:aio_pika.queue:Start to consuming queue: <Queue(queue_in): auto_delete=False, durable=None, exclusive=False, arguments=None>
DEBUG:pika.channel:Adding in on_synchronous_complete callback
DEBUG:pika.channel:Adding passed in callback
Publish b'Hello World: 0'
DEBUG:pika.channel:0 blocked frames
DEBUG:pika.channel:Discarding frame <METHOD(['channel_number=1', 'frame_type=1', "method=<Basic.ConsumeOk(['consumer_tag=ctag1.dce63efa1ef94722a1119231edade9f7'])>"])>
Received (consume callback): b'Hello World: 0'
Publish b'Hello World: 1'
Received (consume callback): b'Hello World: 1'
Publish b'Hello World: 2'
Received (consume callback): b'Hello World: 2'
Publish b'Hello World: 3'
Received (consume callback): b'Hello World: 3'
[...]

PR #59, opened 2017-09-09
Add test_await_get_on_empty_queue_then_publish_and_receive

jakubczaplicki:


jakubczaplicki:
I don't have permissions to push branches here, so I've created this PR from a fork (base: master)

jakubczaplicki:
This test currently fails with TimeoutError because of #56

mosquito:
I already merged your changes, but after that I investigated [how this works in the AMQP](http://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.get) and after that I was have to totally rewrite the test.

PR #62, opened 2017-09-12
Reconnection callback and interval

decaz:
# Rationale
It's useful to have some functional to run something that should be started after the connection established and all objects (channels, queues, exchanges, etc.) were declared in case it was after reconnect.

For instance, when I've had connected, opened channel and declared some queue, I must be sure that my queue is ready and I can use it. So after reconnect this initialization should be run too.

And, of course, reconnection interval as a handy parameter was added :)

### DIFF
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -1,5 +1,4 @@
 import asyncio
-from contextlib import suppress
 from functools import wraps, partial
 from logging import getLogger
 from typing import Callable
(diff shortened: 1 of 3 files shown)


aio_pika/robust_connection.py
```diff
@@ -89,8 +104,7 @@ def close(self) -> None:
         self._closed = True
 
         while self._on_close_callbacks:
-            with suppress(Exception):
```
decaz: @mosquito I guess any exception should be caught within callbacks for explicit.

mosquito: You have to add `finally` block for guarantee to call close.

decaz: @mosquito Yep.

PR #73, opened 2017-10-19
Allow declaring exchanges with 'passive' flag

pierec:
Expose `passive` flag in `aio_pika`.


mosquito:
Great work. Thank you.

PR #83, opened 2017-12-06
Don't create new closing asyncio.Task if already closed in connection __del__

smagafurov:


smagafurov:
Also fix this problem:

```
2017-12-06 14:10:34,565 [ERROR] asyncio: Exception in default exception handler
Traceback (most recent call last):
  File "/Users/sergey/.pyenv/versions/3.6.1/lib/python3.6/asyncio/base_events.py", line 1284, in call_exception_handler
    self.default_exception_handler(context)
  File "/Users/sergey/.pyenv/versions/3.6.1/lib/python3.6/asyncio/base_events.py", line 1256, in default_exception_handler
    value = repr(value)
  File "/Users/sergey/.pyenv/versions/3.6.1/lib/python3.6/asyncio/base_tasks.py", line 15, in _task_repr_info
    coro = coroutines._format_coroutine(task._coro)
  File "/Users/sergey/.pyenv/versions/3.6.1/lib/python3.6/asyncio/coroutines.py", line 274, in _format_coroutine
    assert iscoroutine(coro)
AssertionError
```

Because we create task from task (not coro) in connection.__del__

smagafurov:
@mosquito, may be we should close connection synchronously in `__del__`?

PR #85, opened 2017-12-26
add the x-consistent-hash exchange type

fmang:
RabbitMQ's consistent hash plugin is officially supported and included in the standard distribution. https://github.com/rabbitmq/rabbitmq-consistent-hash-exchange

However, because it's currently missing from the ExchangeType enum, it's awkward to declare such an exchange with with aio-pika. The fix is straightforward.


mosquito:
Thank you for your contribution. I already added this to the aio-pika `1.7.3`.

Issue #75, opened 2017-10-24
Exceptions thrown during reconnect


peter-wimsey:
I am trying to create an AMQP connections that quietly tries to reconnect while the RabbitMQ server is down.

Here's a slightly modified version of the client example:

```python
import asyncio
from aio_pika.robust_connection import RobustConnection

import aio_pika


def on_amqp_lost(connection):
    print('###### on_ampq_lost', connection)


async def ampq(current_loop: asyncio.AbstractEventLoop):
    connection = await aio_pika.connect_robust(
        "amqp://amqp_user:**********@ip-*********.eu-west-1.compute.internal:5672",
        reconnect_interval=10,
        loop=current_loop
    )  # type: RobustConnection

    connection.add_connection_lost_callback(on_amqp_lost)

    async with connection:

        channel = await connection.channel()

        queue = await channel.declare_queue("some_queue", durable=True)

        print('###### connected')

        async for message in queue:
            with message.process():
                print(message.body)


def on_loop_exception(current_loop, context):
    exc = context.get('exception')
    print('###### on_loop_exception %r ' % (exc,))


if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    asyncio.ensure_future(ampq(loop))
    loop.set_exception_handler(on_loop_exception)
    loop.run_forever()
```

This connects fine:

```
###### connected
b'{"message": "hello"}'
```

After restarting the RabbitMQ server (takes about 3 seconds) an _Auto Reconnect Error_ is thrown but the connection seems to recover

```
###### on_ampq_lost amqp://amqp_user:********@ip-*********.eu-west-1.compute.internal:5672
###### on_loop_exception ConnectionError('Auto Reconnect Error',) 
b'{"message": "hello"}'

```

If I shut down the RabbitMQ server I get the last 3 exceptions shown below repeatedly:

```
###### on_ampq_lost amqp://amqp_user:********@ip-*********.eu-west-1.compute.internal:5672
###### on_loop_exception ConnectionError('Auto Reconnect Error',) 
###### on_loop_exception AttributeError("'NoneType' object has no attribute 'channel'",) 
###### on_loop_exception ConnectionError('Auto Reconnect Error',) 
```

Once the RabbitMQ server is up again the connection recovers.

How can I avoid these exceptions? In my use case there are other connections open on the same loop and I want to stop it once any of these disconnects.

mosquito:
I guess this fixed in #88

peter-wimsey:
[...]

PR #88, opened 2018-01-02
Do not attempt to reconnect channels if there is no connection

peter-wimsey:
This seems to fix `AttributeError("'NoneType' object has no attribute 'channel'",)` mentioned in #75.

Would like to add a test to `test_amqp_robust.py` but given that the tests run against a real RabbitMQ server I'm not sure how to synchronise a server restart or similar in the context of a test.


mosquito:
Thank you for your contribution. I already added this to the aio-pika `1.7.2`.

PR #95, opened 2018-01-15
Allow to disable robustness of exchange or queue

decaz:
Within the implementation of my worker class I use robust connection and declare exclusive queue. The queue must be really exclusive within all the worker instances and I correctly handle resource locks for this queue when connecting to the server. But I can't do that when reconnecting because of preliminary repeated declaration of this queue by the `RobustChannel.on_reconnect` implementation. So I suggest to add `robust` parameter for `RobustChannel.declare_exchange` and `RobustChannel.declare_queue` methods to allow disabling robustness of exchange or queue. With this parameters user can customize repeated declaration of any exchange or queue by specific algorithm using reconnect callbacks.


mosquito:
Great work. Thank you.

PR #97, opened 2018-01-18
Fixed auto-naming for instances of RobustQueue

pawelgorzelany:
I'm not sure why this behavior was introduced but it breaks clients that connect to a RabbitMQ server and want to use server-generated queue name.

### DIFF
--- a/aio_pika/robust_queue.py
+++ b/aio_pika/robust_queue.py
@@ -4,8 +4,6 @@ from logging import getLogger
 from types import FunctionType
 from typing import Any, Generator
 
-import shortuuid
-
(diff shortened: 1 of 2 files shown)


mosquito:
It doesn't work when you aio-pika's trying to redefine this queue after reconnection. You can not declare queue with name starts with `amq.` explicitly.

pawelgorzelany:
Ah, that's true. Does it mean that `connect_robust` is not suitable for auto-named queues?

mosquito:
Server-side auto-naming with RobustChannel does not work. Instead of this RobustChannel creates names on the client side.

PR #103, opened 2018-02-02
Ignoring authentication errors within robust connection

decaz:
I suggest not to raise `pika.exceptions.ProbableAuthenticationError` within robust connection and continue retrying to connect.


mosquito:
I think that it is worth at least log this exception.

decaz:
@mosquito I agree. Added error logging.

mosquito:
I already applied this changes to the `2.0.0a0`

PR #110, opened 2018-02-25
Bugfix: _on_getempty should delete _on_getok_callback

dhontecillas:
There is an issue with current implementation: when Pika's ```Channel.basic_get``` method is called, (https://github.com/pika/pika/blob/master/pika/channel.py#L370) checks that the ```_on_getok_callback``` is not set (raising ```DuplicateGetOkCallback``` if it is not None), and then sets the callback.  This callback is cleared in Pika's ```Channel._on_getempty``` and ```Channel._on_getok```. 

**Aio-Pika**'s ```Channel``` implementation overrides ```_on_getempty``` but does not call it's super implementation, leaving the ```_on_getok_callback``` set, so, if the queue is empty, next time we try to use ```Channel.basic_get``` , the ```DuplicateGetOkCallback``` exception is raised.

### DIFF
--- a/aio_pika/adapter.py
+++ b/aio_pika/adapter.py
@@ -101,6 +101,7 @@ class Channel(channel.Channel):
             self._on_getempty_callback(method_frame)
         else:
             LOGGER.debug("Unexcpected getempty frame %r", method_frame)
+        super()_on_getempty(method_frame)
 
     @contextmanager
     def set_get_empty_callback(self, callback):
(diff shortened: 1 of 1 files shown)


mosquito:
Can you write any test which will produce an issue?

dhontecillas:
Will try to find some time to do it this week.

mosquito:
No problem. I already update current test. Great work thank you.

mosquito:
Thanks to you, the new version `aio-pika==2.2.2` is already available.

dhontecillas:
Ok, thank you for finishing my half baked job :)

PR #113, opened 2018-03-07
Use ExternalCredentials when no login provided

MichalJagiello:
That change allow to use ExternalCredentials. Now there is no way to use them


mosquito:
Great work. Thank You.

PR #128, opened 2018-05-31
Handle coroutine wrapping for Python 3.4

Pehat:
Without this fix, you probably get the following error in Python 3.4:
```AttributeError: 'method' object has no attribute '_is_coroutine'```


mosquito:
Great work. Thank you.

Issue #121, opened 2018-04-11
Need to close the channel explicitly?


imnisen:
Hi，When the publisher successfully publish a message, does it need to close the declared channel explicitly?   As shows below:

```
async def publish(data):
       # Here conn is global,made by "connect_robust", which I want to make it long live
        channel = await conn.channel()  
        exchange = await channel.declare_exchange(
            'some_exchange',
            aio_pika.ExchangeType.DIRECT,
            durable=True
        )
        message = aio_pika.Message(
            json.dumps(data).encode(),
            headers={"a": "b")},
            delivery_mode=aio_pika.DeliveryMode.PERSISTENT
        )
        await exchange.publish(message, routing_key="some_key")
        await channel.close()  # Need close here?
```

I want to reuse the connection, so make it global, after that, each time when I need publish a message, I declare the channel, send the message, the close the channel.

My problem is if I close the channel with `await channel.close()`, when call the method `publish` frequently(0.5 second once), this will raise a `pika.exceptions.ChannelClosed`, not each time, several times one exception could happen.
This is very strange, because the channel each time is a new channel, not the closed one last time.

However, if I don't close the channel myself, this works ok.

What cause this problem, or is there another good practise for this.

ps:  aio-pika ==2.3.0, python==3.6.2

Thanks very much!

finetuned89:
I am experiencing this exact issue. After a few messages have been sent, aio-pika will raise an Exception was never retrieved error:

```
Task exception was never retrieved
future: <Task finished coro=<Channel.close() done, defined at /home/gitlab-runner/integration-tests/python_env/lib/python3.6/site-packages/aio_pika/common.py:116> exception=The channel was closed: ()>
Traceback (most recent call last):
  File "/home/gitlab-runner/integration-tests/python_env/lib/python3.6/site-packages/aio_pika/common.py", line 120, in wrap
    raise exceptions.ChannelClosed
pika.exceptions.ChannelClosed

```

It is also impossible to catch this error, if I use:

```
try:
    await my_channel.close()
except Exception as e:
    print("Cannot close channel")  # This is never reached
```

If I don't explicitely close the channels, the number of channels on RabbitMQ keeps growing, which is undesireable. If I do close the channel explicitly, the code keeps working, but the `Task exception was never retrieved` show up.

dan-stone:
The exception is originating from Channel's __del__ method, i.e.

```
# aio-pika/channel.py
# line 335
def __del__(self):
    with suppress(Exception):
[...]

PR #129, opened 2018-05-31
Fixed ChannelClosed exception was never retrieved

finetuned89:
Related to [this issue][https://github.com/mosquito/aio-pika/issues/121].


mosquito:
Excellent. The kind of bugs which I never can to fix by myself. 😅

PR #130, opened 2018-05-31
Added asynchronous context manager support for channels

dan-stone:
Related to #121


aio_pika/channel.py
```diff
@@ -331,9 +347,5 @@ def transaction(self) -> Transaction:
 
         return tx
 
-    def __del__(self):
-        with suppress(Exception):
```
mosquito: Please remove an `import`. This is broke tests.

mosquito:
released in `2.9.0`

PR #135, opened 2018-06-20
Fixed conditions for creating SSL connection.

### LINKED ISSUE #134: connection.connect ignores ssl=True and "amqps" scheme in url
Bug partially introduced in 27b00b2.

Currently to establish SSL connection with amqp url user must provide at least one `ssl_options` field even when defaults are fine.

For example:

```Python
connection = await aio_pika.connect_robust('amqps://guest:guest@localhost:5672', ssl=True)
```

Will create a plain-text connection.

blazewicz:
SSL connection should be created if either of these conditions is met:

* `url` scheme is 'amqps'
* `ssl` argument is set to True
* any `ssl_options` field is set

fixes #134


tests/test_amqp.py
```diff
@@ -1261,3 +1262,17 @@ def test_message_info(self):
         )
 
         self.assertDictEqual(info, msg.info())
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize('url,kwargs,exp', [
+    ('amqps://', {}, {'ssl': True}),
+    ('localhost', {'ssl': True}, {'ssl': True}),
```
blazewicz: if #136 would be merged the 'localhost' could be removed as it is default host name

aio_pika/connection.py
```diff
@@ -394,6 +394,7 @@ def connect(url: str=None, *, host: str='localhost',
         login = url.user or login
         password = url.password or password
         virtualhost = url.path[1:] if len(url.path) > 1 else virtualhost
+        ssl = url.scheme == 'amqps' or ssl
```
mosquito:
```python
ssl = url.scheme == 'amqps' or ssl or ssl_options
```

blazewicz: What I did was intentional to keep the behaviour introduced in https://github.com/mosquito/aio-pika/commit/27b00b21366ab6f0fd09e93b1e0e3b223d884fcf#diff-1bac594ff175e0e7eeccd4a58e3f8269R387

My suggetion is to apply the change you proposed but also remove the filtering of `ssl_options` fields pass it as it is to the `connection_class`

This would let the `connection_class` crash and would give the user useful feedback instead of silently ignoring misspelled options

If you agree I'll edit the PR

blazewicz: That part would be redundant. We already set ssl to True if `ssl_options` is not empty after including parameters from url. No point in checking it twice.

tilsche:
I second this PR. We were just about to write the same PR. Please address this.

aio_pika/connection.py
mosquito: When server's certificate required a client certificate, you have to pass the `ssl_options` parameter, containing certificate and private key paths.

The `ssl=True` or `amqps` schema handling is would be a nice and simple way to enables TLS, but it's should be has lower priority then `ssl_options`.

tilsche: > The ssl=True or amqps schema handling is would be a nice and simple way to enables TLS, but it's should be has lower priority then ssl_options.

Doesn't the PR do exactly that?
https://github.com/mosquito/aio-pika/pull/135/files#diff-1bac594ff175e0e7eeccd4a58e3f8269R414

Using  `amqps://` but not using a client certificate seems to be like a perfectly valid use case, but it's just extremely intuitive without this PR.

PR #136, opened 2018-06-21
fixed inconsistent argument type for connection.connect

blazewicz:
connection.connect is declared to expect 'url' argument as either `str` or `None`
but in tests it was called with `yarl.URL` instance

Also optional argument should tested to be None instead of implicit bool cast. In this case the problem might be that an empty string, which is a valid url, would be ignored by connection.connect.

### DIFF
--- a/aio_pika/connection.py
+++ b/aio_pika/connection.py
@@ -387,8 +387,8 @@ def connect(url: str=None, *, host: str='localhost',
 
     ssl_options = ssl_options or {}
 
-    if url:
-        url = URL(str(url))
+    if url is not None:
+        url = URL(url)
         host = url.host or host
         port = url.port or port
         login = url.user or login
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -42,7 +42,7 @@ class AsyncTestCase(asynctest.TestCase):
 class BaseTestCase(AsyncTestCase):
     @asyncio.coroutine
     def create_connection(self, cleanup=True) -> Generator[Any, None, Connection]:
-        client = yield from connect(AMQP_URL, loop=self.loop)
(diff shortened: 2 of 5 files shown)


aio_pika/connection.py
```diff
@@ -387,8 +387,8 @@ def connect(url: str=None, *, host: str='localhost',
 
     ssl_options = ssl_options or {}
 
-    if url:
-        url = URL(str(url))
+    if url is not None:
+        url = URL(url)
```
mosquito: This operation makes once and the `URL(str(url))` is more comfortable. Yes it's not fast and ugly, but so useful.

tests/__init__.py
```diff
@@ -42,7 +42,7 @@ def get_random_name(self, *args):
 class BaseTestCase(AsyncTestCase):
     @asyncio.coroutine
     def create_connection(self, cleanup=True) -> Generator[Any, None, Connection]:
-        client = yield from connect(AMQP_URL, loop=self.loop)
+        client = yield from connect(str(AMQP_URL), loop=self.loop)
```
mosquito: that's illustrated which the backward compatibility will be broken.

aio_pika/connection.py
blazewicz: casting to str is redundant and removing it wouldn't break compatibilty because yarl.URL can take another object of this type in the constructor and will simply return a copy of it

tests/__init__.py
blazewicz: in declaration of connect() url is annotaded as `str` (actually implicit `Optional[str]`), and only such argument should be provided, otherwise a change in both func spec and documentation is required, but I wouldn't recommend that

```
    :param url: `RFC3986`_ formatted broker address. When :class:`None` \
                will be used keyword arguments.
```

PR #143, opened 2018-07-13
[fix] giant refactoring of pika

mosquito:


decaz:
@mosquito what is the future of such a giant refactoring? :worried:

mosquito:
This is a future https://github.com/mosquito/aiormq, but it not tested as well.

PR #145, opened 2018-08-11
Allow kwargs to be passed to queue.iterator

shanecarey17:
Allow kwargs to be forwarded to call to consume() from call to queue.iterator()


mosquito:
Nice work. Thank you.

mosquito:
This changes released with version `3.1.0`.

PR #156, opened 2018-10-24
Feature/delivery error

mosquito:


decaz:
@mosquito maybe it's worth to make `publish` to behave as it done in pika? I mean to not raise any exceptions but to return `True` or `False`:

https://github.com/pika/pika/blob/e76e2eaeb1ea8b903934dc45bb843ff0fd315fc7/pika/adapters/blocking_connection.py#L2223-L2226

decaz:
@mosquito was my idea inappropriate in this case?

decaz:
@mosquito I agree with you, just noticed that this behavior is different from pika's behavior. When will it be released?

PR #170, opened 2018-12-05
Queue redeclare fails on reconnect if queue is passive

DriverX:
On reconnect passive queues tries redeclared in creation mode and fails over and over again.


DriverX:
Thanks!

Issue #173, opened 2018-12-08
Question: What is the best channel management strategy


krieghan:
I'm struggling with the discovering the best strategy for channel management in a web application that uses aio_pika to publish to a RabbitMQ queue, and I'm hoping that you have some insights into this.  I've looked for information in the documentation, and in other issues for such insights, but I haven't been able to find an authoritative answer to my question.  Here are the options that I've come up with:

1. Create a new channel each time a task needs to publish to the queue, using a context manager.  The channel should be torn down after the task publishes.

(This seems wasteful, as we end up creating a lot of ephemeral channels and the associated exchanges, whose lifetime is very limited)

2. Create a single global channel that every task uses for publishing

3. Create a channel pool (with some min and max number of channels).  Each channel can only be used by one task at a time.

4. Create a channel pool (as above, in 3), but allow each channel to be used by multiple tasks.  The pool should distribute work evenly to all the channels.  Perhaps there should be some ceiling on how many tasks can be publishing on each channel, but perhaps not.

I know that a channel *can* be used by multiple tasks at the same time.  I think the main thing that I do not understand is if there is some performance benefit of having multiple channels, or some maximum number of communications that a channel can be engaged in simultaneously.  Documentation for Rabbit and AMQP and various discussions about channels seem to revolve around having one channel-per-thread, but I'm not quite sure how that applies to use in aio applications.

mosquito:
@krieghan I preparing to merge #174. When you have any comments feel free to write it there.

krieghan:
Wow, I didn't expect an actual implementation.  Thanks!

PR #174, opened 2018-12-11
[fix] add abstract pool

mosquito:
Inspired by #173 from @krieghan.

### DIFF
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from typing import AsyncContextManager, TypeVar, Coroutine, Callable, Any
+from typing import Any, AsyncContextManager, Callable, Coroutine, TypeVar
 
 
 T = TypeVar("T")
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -30,14 +30,16 @@ class Pool:
         return PoolItemContextManager(self)
 
     @property
-    def _is_overflow(self):
+    def _has_released(self):
+        return self.__items.qsize() > 0
+
+    @property
+    def _is_overflow(self) -> bool:
         if self.__max_size:
-            return self.__created >= self.__max_size
+            return self.__created >= self.__max_size or self._has_released
+        return self._has_released
 
-    async def _get(self) -> T:
-        if self._is_overflow:
-            return await self.__items.get()
-
+    async def _create_item(self) -> T:
         async with self.__lock:
             if self._is_overflow:
                 return await self.__items.get()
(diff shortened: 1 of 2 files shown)


aio_pika/pool.py
```diff
@@ -0,0 +1,66 @@
+import asyncio
+import logging
+from typing import AsyncContextManager, TypeVar, Coroutine, Callable, Any
```
alvassin: wrong imports order

tests/test_pool.py
```diff
@@ -0,0 +1,57 @@
+import asyncio
+
+import pytest
+from aio_pika.pool import Pool
+from tests import AsyncTestCase
```
alvassin: Why not used fixtures?

aio_pika/pool.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+T = TypeVar("T")
+ItemType = Coroutine[Any, None, T]
+ConstructorType = Callable[[], ItemType]
```
leenr: `Awaitable[ItemType]`?
> `item = await self.__constructor()`

aio_pika/pool.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, pool: Pool):
+        self.pool = pool
+        self.item = None
+
+    async def __aenter__(self) -> T:
+        self.item = await self.pool._get()
+        return self.item
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        self.pool.put(self.item)
```
leenr: `self.item = None`?

aio_pika/pool.py
```diff
@@ -0,0 +1,66 @@
...
+    __slots__ = (
+        'loop', '__max_size', '__items',
+        '__constructor', '__created', '__lock'
+    )
+
+    def __init__(self, constructor: ConstructorType, max_size: int = None,
+                 loop: asyncio.AbstractEventLoop = None):
+
+        self.loop = loop or asyncio.get_event_loop()
+        self.__max_size = max_size or -1
```
leenr: It seems that `-1` is not supported by the code below.
If `self.__max_size == -1`, `self._is_overflow` will be always `True`, and `await self.__items.get()` will hang in the `_get`.

tests/test_pool.py
mosquito: All another test suite uses Unittest

krieghan:
I tested this on my application under a pretty heavy load test.  It works without issues, and out-performs the pool that I'd thrown together.

PR #176, opened 2018-12-16
Create py.typed (add PEP 561 support)

zarybnicky:
Aio_pika uses some type annotations internally, but packages depending on aio_pika that use e.g. can't use those without either adding interface files to python/typeshed, or without adding a py.typed file to the root if the main package. This PR does the second thing, adds an empty py.typed file.


zarybnicky:
On further reading, this file should contain `partial\n` and it should be packaged via package_data in setup.py, so I'll have to get back to this later today.

PR #183, opened 2019-01-13
Pool: use None instead of Elipsis for initial state

chibby0ne:
Usage of None is standard on Python for uninitialized objects, while Ellipsis is mostly used as a helper for slicing multidimensional lists/arrays.
Using None provides clarity and simplicity.

Signed-off-by: Antonio Gutierrez <chibby0ne@gmail.com>

### DIFF
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -73,4 +73,3 @@ class PoolItemContextManager(AsyncContextManager):
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         if self.item is not None:
             self.pool.put(self.item)
- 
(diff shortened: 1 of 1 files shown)


mosquito:
Yes looks as an simplification, but that's an abstract pool. Theoretically the returning constructor value might be None. In this case constructor will be called every time. It's a real reason for using Elipsis as an initial value.

chibby0ne:
Thanks for your quick reply @mosquito!
I understand what you mean. The constructor would be called every time but it would not be stored in the Queue.
The only thing is I assumed that the constructor could only return a Coroutine according to https://github.com/mosquito/aio-pika/blob/ea7ad14143f75275cb9785deaf3df61b0041e9a8/aio_pika/pool.py#L7-L8. Is my assumption wrong?

PR #184, opened 2019-01-13
pool: enable arguments for pool constructor

chibby0ne:
Having arguments enables flexible and reusable pool creation functions.

Signed-off-by: Antonio Gutierrez <chibby0ne@gmail.com>

### DIFF
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -5,7 +5,7 @@ from typing import Any, AsyncContextManager, Callable, Coroutine, TypeVar
 
 T = TypeVar("T")
 ItemType = Coroutine[Any, None, T]
-ConstructorType = Callable[[], ItemType]
+ConstructorType = Callable[..., ItemType]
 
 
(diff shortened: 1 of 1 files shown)


mosquito:
Thank you for your contribution. I made similar changes before release, but "channel creating" example brake this idea. For passing function with args and kwargs as a constructor the better way is using `partial` class.

chibby0ne:
> I made similar changes before release, but "channel creating" example brake this idea. 

Right. I was planning on using the arguments only for the creation of connections, which might vary on the hostname, port, user/pass, etc. Creation of channel on the other hand doesn't really need any parameters, that's why it would receive an empty set `()` seen in: https://github.com/mosquito/aio-pika/blob/bdf3f9d231dca3483a214cc7899af52aa3547c09/aio_pika/pool.py#L27

> For passing function with args and kwargs as a constructor the better way is using partial class.

Thanks for the tip. Yep you're right. I will do the change and submit it again.

Issue #124, opened 2018-04-16
AsyncioConnection blocking during getaddrinfo


steersbob:
## Problem

When attempting to connect to a non-existing host, `aio_pika.connect_robust()` blocks the application thread, and periodically logs critical messages.

## Version

aio-pika: 2.6.0
pika: 0.10.0

## How to reproduce

```python
import asyncio
import aio_pika

async def talk():
    while True:
        print('something')
        await asyncio.sleep(0.5)

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(
        asyncio.wait(
            [
                aio_pika.connect_robust("amqp://guest:guest@NOT_A_HOST", loop=loop),
                talk()
            ]
        )
    )
    loop.close()
```

We'd expect this to print 'something' every 500ms (twice per default reconnect interval), in between connection errors.

Actual output (interval much longer than 1 second):
```
Connection refused: 500 - [Errno -2] Name or service not known
something
Connection refused: 500 - [Errno -2] Name or service not known
something
Connection refused: 500 - [Errno -2] Name or service not known
```

## Root Cause

`AsyncioConnection(base_connection.BaseConnection)` in `aio_pika/adapter.py` does not replace all blocking I/O functions used by its superclass.

Code trace (`aio_pika/adapter.py`):

```python
def _adapter_connect(self):
    error = super()._adapter_connect()
    ...
```

calls `pika/adapters/base_connection.py`

```python
def _adapter_connect(self):
    ...
    while True:
        try:
            addresses = socket.getaddrinfo(self.params.host, self.params.port,
                                                               0, socket.SOCK_STREAM,
                                                               socket.IPPROTO_TCP)
```

## Possible solutions

Disclaimer: I'm not a fan of any of these, but they'd technically solve the problem.

* Monkey patch `socket.getaddrinfo` to use `asyncio.BaseEventLoop.getaddrinfo` instead.
* Duplicate the `BaseConnection._adapter_connect()` code into `AsyncioConnection._adapter_connect()`, avoiding a call to super.
* Do a separate `loop.getaddrinfo()` call loop in `AsyncioConnection._adapter_connect()`, to assure that the blocking function in its superclass will only be called once when success is guaranteed.

dequis:
[...]

PR #186, opened 2019-01-14
Make pika connections async, move _adapter_connect to a thread pool

dequis:
Laziest approach, this code could really leverage asyncio's high level connection stuff.

----

For #124 

This is mostly just a RFC, I don't expect this to get merged, but it works as expected and i'm fairly sure it's thread safe.

The most important change here, regardless of the approach taken, is to remove the connect() call from `pika.connection.Connection.__init__`, which can never be async, and to move it to an explicit call after creating the instance, so that it can be awaited.

I went for minimal but not quite API-clean changes - I don't know what API guarantees should be kept here, or what's the desired extent of changes to the vendored pika, if any.

At first I avoided changing the vendored pika but later I realized its BaseConnection is nothing like the upstream one, so I simplified things. In an [earlier implementation of this change][1] I didn't touch pika at all but had to override `connect()` with a no-op and copy a whole method.

[1]: https://github.com/dequis/aio-pika/commit/f531a1aea51a6c9a76acb86db9cb30dcaa2e9972

### DIFF
--- a/aio_pika/adapter.py
+++ b/aio_pika/adapter.py
@@ -136,8 +136,8 @@ class AsyncioConnection(base_connection.BaseConnection):
                          on_close_callback, self.ioloop,
                          stop_ioloop_on_close=stop_ioloop_on_close)
 
-    def _adapter_connect(self):
-        error = super()._adapter_connect()
+    async def _adapter_connect(self):
+        error = await self.loop.run_in_executor(None, self._sync_adapter_connect)
 
         if not error:
             self.ioloop.add_handler(
--- a/aio_pika/adapter.py
+++ b/aio_pika/adapter.py
@@ -137,7 +137,9 @@ class AsyncioConnection(base_connection.BaseConnection):
                          stop_ioloop_on_close=stop_ioloop_on_close)
 
     async def _adapter_connect(self):
-        error = await self.loop.run_in_executor(None, self._sync_adapter_connect)
+        error = await self.loop.run_in_executor(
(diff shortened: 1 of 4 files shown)


mosquito:
Version 4.9.3 contains this changes. Thank you.

dequis:
Huh, that was unexpected, but thanks!

What do you think about moving forward in terms of using more native asyncio socket APIs? Say, using the higher-level `loop.create_connection` (and Protocol/Transport) or even keep similar code but using `loop.sock_connect`/`loop.getaddrinfo` I think a lot of code could be made redundant with the former, but I'm not sure if it's okay to add asyncio loop calls to `BaseConnection`, where most of the blocking calls live.

Also I might still be running into main-loop-blocking code somewhere else, but need to debug some more to know where to apply an equivalent fix.

mosquito:
I planning to remove pika's code from aio-pika. Part of **5.x.x** roadmap is switching to the [aiormq](https://github.com/mosquito/aiormq) instead of pika, but aio-pika is a production ready solution and this plan is quite risky. 

Therefore volunteers who can take the risk and test aiormq for production project are needed.

pros:
* resolves all known performance issues
* pika's connection creating code takes almost half of pika's connector class code but is no actually needed anymore (see. [asyncio.open_connection](https://docs.python.org/3/library/asyncio-stream.html#asyncio.open_connection)).
* the "publisher confirms" tracking is already moved to the aiormq.

cons:
* possible breaking API changes
* possible tests incompatibility.

PR #195, opened 2019-02-15
Fix HeaderProxy bug

tzoiker:
`HeaderProxy` has incorrect iterator behaviour, which breaks `Message.__copy__`.


mosquito:
Thank you.

PR #198, opened 2019-02-23
Fix encode timestamp error on copy

tzoiker:
`encode_timestamp` fails on `time.time_struct` value


mosquito:
Could you please add some tests for prevent future regression?

tzoiker:
https://github.com/mosquito/aio-pika/pull/198/files#diff-3c7fac459f9ff713e30df0d7d0ae00ceR1414

PR #212, opened 2019-04-05
Allow to specify `requeue=` and `reject_on_redelivered=` in Master pattern

leenr:
Otherwise, there is no way to specify `reject_on_redelivered=True` when using `Master` pattern.


mosquito:
Released in `aio-pika==5.5.1`

PR #222, opened 2019-05-08
fix return type of Queue.bind

tilsche:
During debugging in stumbled on an odd return type of `Queue.bind` that seems to be a typo.

This appears to be inconsequential as the type is not checked by default.

Note that there is an actual runtime check in `Channel.rpc` for `valid_responses` and the return type is in fact `BindOk`.

I have not actually tested the patch as there isn't a test case where it fails.


mosquito:
I am back from my vacation, and now I return to the accumulated requests and issues. Thanks for your contribution.

PR #234, opened 2019-08-08
Fix robust connection

decaz:
There is unexpected `return` statement within connection loop so `connect` method will return after the first iteration.

### DIFF
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -4,7 +4,7 @@ from logging import getLogger
 from typing import Callable
 
 from aiormq.connection import parse_bool, parse_int
-from .exceptions import ChannelClosed, AMQPError
+from .exceptions import AMQPError, CONNECTION_EXCEPTIONS
 from .connection import Connection, connect
 from .robust_channel import RobustChannel
 
@@ -40,7 +40,6 @@ class RobustConnection(Connection):
         self.fail_fast = self.kwargs['fail_fast']
 
         self.__channels = set()
-        self._on_connection_lost_callbacks = set()
         self._on_reconnect_callbacks = set()
         self._closed = False
 
@@ -48,7 +47,7 @@ class RobustConnection(Connection):
     def _channels(self) -> dict:
         return {ch.number: ch for ch in self.__channels}
 
-    def _on_connection_close(self, connection, closing, **_):
+    def _on_connection_close(self, connection, closing, *args, **kwargs):
         self.connection = None
 
         # Have to remove non initialized channels
@@ -72,13 +71,13 @@ class RobustConnection(Connection):
         self._on_reconnect_callbacks.add(callback)
 
     async def connect(self, timeout=None):
-        if self.fail_fast:
-            return await super().connect(timeout=timeout)
-
         while True:
             try:
-                await super().connect(timeout=timeout)
-            except (ConnectionError, AMQPError):
+                return await super().connect(timeout=timeout)
+            except CONNECTION_EXCEPTIONS:
+                if self.fail_fast:
+                    raise
+
                 log.warning(
                     "First connection attempt failed "
                     "and will be retried after %d seconds",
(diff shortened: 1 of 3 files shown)


aio_pika/robust_connection.py
```diff
@@ -72,13 +72,13 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
     async def connect(self, timeout=None):
-        if self.fail_fast:
-            return await super().connect(timeout=timeout)
-
         while True:
             try:
-                await super().connect(timeout=timeout)
-            except (ConnectionError, AMQPError):
+                return await super().connect(timeout=timeout)
+            except:
```
decaz: @mosquito I use broad exception here because there are `pamqp` exceptions can be raised for instance when connecting while RabbitMQ is starting. Unfortunately `pamqp.specification` exceptions aren't inherited from base `PAMQPException` so we should do something like this if broad exception is undesirable:
```python
except (
    ConnectionError,
    AMQPError,
    pamqp.exceptions.PAMQPException
) + tuple(pamqp.specification.ERRORS.values()):
```

aio_pika/robust_connection.py
```diff
@@ -72,13 +72,13 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
     async def connect(self, timeout=None):
-        if self.fail_fast:
-            return await super().connect(timeout=timeout)
-
         while True:
             try:
-                await super().connect(timeout=timeout)
-            except (ConnectionError, AMQPError):
+                return await super().connect(timeout=timeout)
+            except (ConnectionError, AMQPError) + PAMQP_EXCEPTIONS:
```
mosquito: Could you please define this tuple in global scope?

aio_pika/robust_connection.py
```diff
@@ -124,7 +122,7 @@ def channel(self, channel_number: int = None,
         for number, channel in self._channels.items():
             try:
                 await channel.on_reconnect(self, number)
-            except (RuntimeError, ChannelClosed, AMQPError):
+            except (RuntimeError, ChannelClosed) + CONNECTION_EXCEPTIONS:
```
decaz: I guess this is not so perfect. @mosquito what do you think?

aio_pika/robust_connection.py
decaz: Done.

aio_pika/robust_connection.py
mosquito: The global variable in module it's free I think. But creating a new tuple when preparing exception block isn't free.

decaz: I returned changes back without `ChannelClosed` exception which inherits from `AMQPError`.

aio_pika/robust_connection.py
```diff
@@ -124,7 +121,7 @@ def channel(self, channel_number: int = None,
         for number, channel in self._channels.items():
             try:
                 await channel.on_reconnect(self, number)
-            except (RuntimeError, ChannelClosed, AMQPError):
+            except (RuntimeError, AMQPError):
```
decaz: @mosquito should we check for `CONNECTION_EXCEPTIONS` here? I mean `ConnectionError` and `PAMQP_EXCEPTIONS`.

decaz: @mosquito ?

mosquito:
All changes for `5.x.x` was merged and tested. But this changes propose different behaviour for `RobustConnection` logic, thats why major version `aio-pika==6.0.0` was bumped and already released. 

@decaz thank you for your contribution in this library. Your explicit feedback very important for me.

decaz:
@mosquito these changes don't bring different behavior, - they just fix `RobustConnection` in case `fail_fast='0'`. Maybe it worth to merge this changes into `5.x.x`?

mosquito:
Looks risky. Why you want to stay on 5.x.x?

decaz:
I am not tied to version 5, these changes just fix it without introducing backward incompatibility. However, it’s too late to discuss it, because version 6 is already released :)

PR #237, opened 2019-08-18
Extended ExchangeType

daniel55411:
Hello, 
I think the type exchange restriction is strong. Sometimes you need exchange type which haven't been added yet. For example, has been created new plugin and __aio_pika__ has not been updated yet

### DIFF
--- a/aio_pika/exchange.py
+++ b/aio_pika/exchange.py
@@ -19,6 +19,7 @@ class ExchangeType(Enum):
     HEADERS = 'headers'
     X_DELAYED_MESSAGE = 'x-delayed-message'
     X_CONSISTENT_HASH = 'x-consistent-hash'
+    X_MODULUS_HASH = 'x-modulus-hash'
 
 
 class Exchange:


aio_pika/exchange.py
```diff
@@ -19,13 +20,15 @@ class ExchangeType(Enum):
...
     X_DELAYED_MESSAGE = 'x-delayed-message'
     X_CONSISTENT_HASH = 'x-consistent-hash'
+    X_MODULUS_HASH = 'x-modulus-hash'
 
 
 class Exchange:
     """ Exchange abstraction """
 
     def __init__(self, connection, channel: aiormq.Channel, name: str,
-                 type: ExchangeType = ExchangeType.DIRECT, *,
```
mosquito: the [same](https://github.com/mosquito/aio-pika/blob/master/aio_pika/types.py#L8)

mosquito:
The exchange type passing to the `aiormq.Channel` and will be cast as `str` there.

daniel55411:
Oh... sorry for wasting your time, but could you suggest how add new exchange type if that type has not been added yet in __aio_pika__?

mosquito:
Library doesn't restrict string arguments for exchange types. Enums was added just for usability and your fell free fo inherit this enum and use it as an argument. When it's doesn't work it's my problem and you should create an issue :-)

In practice RabbitMQ extensible around special Exchanges, and only Exchange has no strict Enum-values restriction.

mosquito:
Really my fail is lack of note in the documentation or example with demonstrate solution of this problem.

daniel55411:
Thank you very much!
Sorry if I misunderstood you, but I have already tried to extend `ExchangeType`
There is output of python console:
```python
>>> import aio_pika
>>> class ExtendedExchangeType(aio_pika.ExchangeType):
...          X_MODULUS_HASH = 'x-modulus-hash'
...     
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/usr/lib/python3.7/enum.py", line 130, in __prepare__
    member_type, first_enum = metacls._get_mixins_(bases)
  File "/usr/lib/python3.7/enum.py", line 471, in _get_mixins_
    raise TypeError("Cannot extend enumerations")
TypeError: Cannot extend enumerations
```

And I have tried declare exchange:
```python
>>> async def test():
...     connection = await aio_pika.connect()
...     channel = await connection.channel()
...     return await channel.declare_exchange('test', type='x-modulus-hash')
...
>>> import asyncio
>>> loop = asyncio.get_event_loop()
>>> loop.run_until_complete(test())
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/usr/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "<input>", line 4, in test
  File "/home/daniel/Projects/aio-pika/aio_pika/channel.py", line 216, in declare_exchange
    internal=internal, passive=passive, arguments=arguments
  File "/home/daniel/Projects/aio-pika/aio_pika/exchange.py", line 40, in __init__
    self.__type = type.value
AttributeError: 'str' object has no attribute 'value'
```
I apologize again if I did not understand you!

PR #238, opened 2019-08-22
 Make RobustConnection more robust

decaz:
* Added `RuntimeError` to `CONNECTION_EXCEPTIONS`
* Handle `CONNECTION_EXCEPTIONS` at channel reconnect

### DIFF
--- a/aio_pika/exceptions.py
+++ b/aio_pika/exceptions.py
@@ -20,10 +20,15 @@ from aiormq.exceptions import (
     ChannelNotFoundEntity
 )
 
-PAMQP_EXCEPTIONS = (pamqp.exceptions.PAMQPException,) + tuple(
-    pamqp.specification.ERRORS.values()
-)
-CONNECTION_EXCEPTIONS = (ConnectionError, AMQPError) + PAMQP_EXCEPTIONS
+PAMQP_EXCEPTIONS = (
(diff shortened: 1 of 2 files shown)


aio_pika/connection.py
```diff
@@ -52,12 +52,12 @@ def __init__(self, url, loop=None, **kwargs):
         self.kwargs = self._parse_kwargs(kwargs or self.url.query)
 
         self.connection = None     # type: aiormq.Connection
-        self.close_callbacks = set()
+        self._close_callbacks = set()
```
mosquito: why you want to hide this? User may use it directly It's legal.

decaz: It's the only place where callbacks set is public, - all other callback sets are private. I've made this to uniform callbacks storages.

mosquito: 🤔

decaz: @mosquito what do you think?

mosquito: Uniform callback storage sounds good. Let's do it.

mosquito:
@decaz my vision of uniform callback api requested in #239

decaz:
@mosquito I've removed all changes of callback usage in favor of #239.

PR #239, opened 2019-08-22
CallbackCollection

mosquito:


aio_pika/tools.py
```diff
@@ -63,3 +64,62 @@ def wrap(*args, **kwargs):
...
+
+    def freeze(self):
+        if self.is_frozen:
+            raise RuntimeError("Collection already frozen")
+
+        self.__callbacks = frozenset(self.__callbacks)
+
+    def unfreeze(self):
+        if not self.is_frozen:
+            raise RuntimeError("Not collection is not frozen")
```
decaz: Redundant "Not".

aio_pika/channel.py
```diff
@@ -65,6 +66,14 @@ def __init__(self, connection, channel_number: int = None,
         # noinspection PyTypeChecker
         self.default_exchange = None       # type: Exchange
 
+    @property
+    def done_callbacks(self) -> CallbackCollection:
```
decaz: Why do you wrap callback collections in properties?

decaz:
@mosquito looks pretty complicated for simple storage.. Maybe it worth to leave all callback sets but make them "public" and not to embed such kind of type as `CallbackCollection` (keep it simple and let users do whatever they want with callback sets) ?

aio_pika/channel.py
mosquito: for restrict setter behaviour

mosquito:
@decaz in my case freezing the callback collection it is what I really want.
The second problem is a simple set do not checks callback before add it.

aio_pika/tools.py
```diff
@@ -1,8 +1,9 @@
 import asyncio
 from functools import wraps
+from collections import Set
```
decaz: collections.abc (to avoid `DeprecationWarning`)

decaz:
I think something like `__call__` method could be added to `CallbackCollection` to run all the callbacks in collection instead of duplicating.

aio_pika/tools.py
decaz: @mosquito FYI

decaz:
@mosquito what do you think about `CallbackCollection.__call__` method?

mosquito:
@decaz nice idea, I will make it on separate PR

PR #245, opened 2019-09-03
Add missing connection timeout parameter

decaz:
Some type hints were also fixed.

### DIFF
--- a/aio_pika/connection.py
+++ b/aio_pika/connection.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 from functools import partial
-from typing import Callable, Type
+from typing import Callable, Type, TypeVar
 
 from yarl import URL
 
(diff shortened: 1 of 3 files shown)


mosquito:
Looks so complex against `asyncio.wait_for(connect(), timeout=1)`

decaz:
@mosquito there is already at low-level:
https://github.com/mosquito/aio-pika/blob/83bf56db507a8a88eb80dce353067c608134a76a/aio_pika/connection.py#L116-L119
but high-level `connect` and `connect_robust` doesn't have `timeout` parameter to propagate timeout to the lower level, - that is what this PR for.

mosquito:
@decaz you are right.

mosquito:
version 6.1.1 already released

PR #247, opened 2019-09-11
max reconnect attempts

harlov:
PR adds a limit for reconnect attempts. It's useful in case then connect parameters have changed and we need gracefully shutdown and re-init client.


mosquito:
@harlov please rebase this branch against master.

aio_pika/exceptions.py
```diff
@@ -40,6 +40,10 @@ class QueueEmpty(AMQPError, asyncio.QueueEmpty):
     pass
 
 
+class MaxReconnectAttemptsReached(Exception):
```
dizballanze: Maybe, this class should be a subclass of `ConnectionError`

roperto:
Any interest in moving this one ahead? I can help rebasing/testing if needed, but only if maintainers have interest in this feature.

In my use case I need to stop a service if I lose connection (it will respawn later with correct parameters). But anything around adjusting the URL to reconnect could be useful too.

PR #249, opened 2019-09-12
Add robust parameters

decaz:
As it is already done for `RobustChannel.declare_exchange` and `RobustChannel.declare_queue`.

### DIFF
--- a/aio_pika/robust_exchange.py
+++ b/aio_pika/robust_exchange.py
@@ -42,15 +42,16 @@ class RobustExchange(Exchange):
             await self.bind(exchange, **kwargs)
 
     async def bind(self, exchange, routing_key: str='', *,
-                   arguments=None, timeout: int=None):
+                   arguments=None, timeout: int=None, robust: bool = True):
         result = await super().bind(
             exchange, routing_key=routing_key,
             arguments=arguments, timeout=timeout
         )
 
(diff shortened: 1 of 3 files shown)


decaz:
@mosquito any thoughts?

mosquito:
I want to add some tests for this but not imagine how.

decaz:
@mosquito maybe it can be merged for now as it was done at #95 ?

mosquito:
please add following test case

```python
class TestCaseNoRobust(AMQPTestCase):
    async def create_connection(self, cleanup=True) -> Connection:
        client = await connect_robust(str(AMQP_URL), loop=self.loop)
        if cleanup:
            self.addCleanup(client.close)
        return client

    async def create_channel(self, connection=None,
                             cleanup=True, **kwargs) -> Channel:
        return await super().create_channel(
            cleanup=cleanup, robust=False
        )

    async def declare_queue(self, *args, **kwargs) -> Queue:
        kwargs = kwargs.copy()
        kwargs['robust'] = False
        return await super().declare_queue(*args, **kwargs)

    async def declare_exchange(self, *args, **kwargs) -> Exchange:
        kwargs = kwargs.copy()
        kwargs['robust'] = False
        return await super().declare_exchange(*args, **kwargs)
```

decaz:
@mosquito done.

PR #259, opened 2019-10-03
Add global timeout for all operations within connection

decaz:
Currently every operation should be called with `timeout` parameter to catch timeout of operations. With `operation_timeout` parameter of connection it would be much easier.


decaz:
Default value of `timeout` parameters should be some special object (`OPERATIONS_TIMEOUT` for example) because `None` should mean no timeout at all. Will add it soon.

tests/test_amqp.py
```diff
@@ -1071,7 +1071,7 @@ def share(*a, **kw):
 
         await queue.bind(exchange, routing_key)
 
-        body = bytes(shortuuid.uuid(), 'utf-8') * 9999999
+        body = bytes(shortuuid.uuid(), 'utf-8') * 6000000
```
mosquito: Why it’s affected?

decaz: I've mentioned it in commit body: `Since RabbitMQ 3.8.0 max size of message is 128MB.` Otherwise test fails.

mosquito:
Please rebase

decaz:
@mosquito done.

mosquito:
Sorry for long answer, but why you want to add a parameter to the methods and use `asyncio.wait_for` inside them instead of wrapping `asyncio.wait_for` externally?

decaz:
@mosquito because `wait_for` and `timeout` parameters are already inside library methods and you don't have to wrap them everywhere within your code.

aviramha:
Update on this?

mosquito:
please resolve conflicts

Issue #260, opened 2019-10-04
Type not converted from str to ExchangeType in declare_exchange


heckad:
```
connection = await aio_pika.connect(
    rabbitmq_option('url'),
)

async with connection.channel() as channel:
    exchange = await channel.declare_exchange('new', 'fanout', durable=True)
```

mosquito:
#262 should fix it.

heckad:
Yes. #262 fix it.

PR #261, opened 2019-10-04
[fix] Not converted type

heckad:
https://github.com/mosquito/aio-pika/issues/260

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -210,6 +210,8 @@ class Channel:
 
         if auto_delete and durable is None:
             durable = False
+        
+        type = ExchangeType(type)
 
         exchange = self.EXCHANGE_CLASS(
             connection=self._connection, channel=self.channel,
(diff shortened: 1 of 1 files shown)


aio_pika/channel.py
```diff
@@ -210,6 +210,8 @@ def _on_return(self, message: aiormq.types.DeliveredMessage):
 
         if auto_delete and durable is None:
             durable = False
+        
+        type = ExchangeType(type)
```
mosquito: This was done intentionally. The fact is that in RabbitMQ, extensions use custom Exchange types other than this enumeration. So this check breaks the use of third-party Exchange types.

heckad: Then you have the wrong type hint. I tried to send a string and got an error. Look [#260](https://github.com/mosquito/aio-pika/issues/260).

heckad:
Why did tests fail in different places? My fix can't break it.

aio_pika/channel.py
heckad: @mosquito ?

mosquito:
Please rebase against master

PR #263, opened 2019-10-09
Add reopen method for channel

decaz:
It happens that the channel closes and you need to open it again. In case of robust connection currently this is possible using an inconvenient call:
```python
await channel.on_reconnect(
    connection=channel._connection,
    channel_number=channel._channel_number,
)
```
I suggest to make it as follows:
```python
await channel.reopen()
```
Thereby `on_reconnect` methods should be renamed to `restore` which mean more broad sense than just reconnection.

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -186,6 +186,10 @@ class Channel:
     def _on_return(self, message: aiormq.types.DeliveredMessage):
         self._return_callbacks(IncomingMessage(message, no_ack=True))
 
+    async def reopen(self):
+        self._channel = None
+        await self.initialize()
+
     async def declare_exchange(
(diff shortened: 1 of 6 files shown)


mosquito:
@decaz you have to write some tests for reopening a closed channel

mosquito:
But tests 😟

decaz:
@mosquito I didn't requested a review yet :) Just rebased againt the master branch.

mosquito:
Ok. Looking forward tests

decaz:
@mosquito I've added a little test. What do you think?

decaz:
@mosquito any thoughts?

mosquito:
Hi there. Current plan is make a new release after new year holidays. I'm so busy sorry.

Issue #200, opened 2019-02-25
Callbacks were gone


decaz:
There are no more `add_connection_lost_callback` and `add_reconnect_callback` callbacks for connection.

mosquito:
`add_reconnect_callback` still [here](https://github.com/mosquito/aio-pika/blob/master/aio_pika/robust_connection.py#L66) and available only for` RobustConnection`.

`add_connection_lost_callback` it's a lost breaking change from `CHANGELOG.md`. Now it's [`add_close_callback`](https://github.com/mosquito/aio-pika/blob/master/aio_pika/connection.py#L67) method available for both connection types.

decaz:
Ok, thanks!

PR #266, opened 2019-10-30
Add `global_` parameter to `Channel.set_qos` method

### LINKED ISSUE #169: all_channels at the RobustChannel's QoS
What's the reason of unavailability of `all_channels` parameter at the `RobustChannel.set_qos` method?

https://github.com/mosquito/aio-pika/blob/57af0c3f3f81f49e9593f7d96ef624582ce3be79/aio_pika/robust_channel.py#L87-L90

decaz:
Closes #169.

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -1,6 +1,7 @@
 import asyncio
 from enum import unique, Enum
 from logging import getLogger
+from warnings import warn
 
 from aio_pika.tools import CallbackCollection
 
@@ -263,8 +264,12 @@ class Channel:
 
     async def set_qos(
         self, prefetch_count: int = 0, prefetch_size: int = 0,
-        global_: bool = False, timeout: TimeoutType = None
+        global_: bool = False, timeout: TimeoutType = None,
+        all_channels: bool = None
     ) -> aiormq.spec.Basic.QosOk:
+        if all_channels is not None:
+            warn('Use "global_" instead of "all_channels"', DeprecationWarning)
+            global_ = all_channels
 
         return await asyncio.wait_for(
             self.channel.basic_qos(
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -263,13 +263,14 @@ class Channel:
 
     async def set_qos(
         self, prefetch_count: int = 0, prefetch_size: int = 0,
-        all_channels: bool = False, timeout: TimeoutType = None
+        global_: bool = False, timeout: TimeoutType = None
     ) -> aiormq.spec.Basic.QosOk:
 
         return await asyncio.wait_for(
             self.channel.basic_qos(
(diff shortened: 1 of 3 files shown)


mosquito:
passing kwargs will be better and backward compatible.

decaz:
What kind of backward compatibility do you mean?

If you mean remove of `all_channels` parameter then I think nobody used it because it wasn't used inside of library in any way and didn't gave any effect. Can we can just remove it to not to bring some complexity in such a simple case?

If you mean incompatibility with `aiormq` then we can change minimal required version of it.

What do you think?

mosquito:
1. Yes! I want to restrict `aiormq>3,<4`.
2. `all_channels` might be used by 4.x.x version users.

mosquito:
`all_channels` actually means `global_` in this case.
I see two ways:
1. marking `all_channels` deprecated and add `global_` as an alias for `all_channels`.
2. just add alias for `all_channels` to `global_`

decaz:
Why do you want to save parameter that was used by **previous major** version? There were another breaking changes at 5 version (for example https://github.com/mosquito/aio-pika/issues/200#issuecomment-467613597). And now `all_channels` has no effect at all and if someone is using it within 5th version then he will have to just remove it and that's all.

aio_pika/channel.py
```diff
@@ -263,13 +263,14 @@ def _on_return(self, message: aiormq.types.DeliveredMessage):
 
     async def set_qos(
         self, prefetch_count: int = 0, prefetch_size: int = 0,
-        all_channels: bool = False, timeout: TimeoutType = None
```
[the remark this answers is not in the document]
decaz: @mosquito fixed.

setup.py
```diff
@@ -38,7 +38,7 @@
     packages=find_packages(exclude=['tests']),
     package_data={'aio_pika': ['py.typed']},
     install_requires=[
-        'aiormq>=2.9,<3',
+        'aiormq>=3,<4',
```
mosquito: please add `>=3.0.1,<4`

mosquito: Or just rebase against master

decaz:
@mosquito done.

mosquito:
Clean work. But I'm not sure about `RobustConnection` behaviour. Might be better to ignore it for `RobustChannel`?

decaz:
Why do we need to change behavior if connection is robust? If it's robust then how can channel QoS be global among multiple consumers?

mosquito:
After reconnection new connection has no had globally set QoS. And connection state will be different after reconnecting. The correct way is propagating last set QoS with the `global_` parameter into `RobustConnection` from `RobustChannel` as a connection property.

mosquito:
Please rebase against the master and I hope it will be merged soon.

decaz:
@mosquito rebase done.

PR #267, opened 2019-11-13
Stucking RobustConnection

mosquito:
Fixes stuck RobustConnectionwhen a queue has been declared with the exclusive flag.

### DIFF
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -32,7 +32,6 @@ class Pool:
         self.__constructor_args = args or ()
         self.__created = 0
         self.__lock = asyncio.Lock()
-        self.__close_lock = asyncio.Lock()
 
     def acquire(self) -> 'PoolItemContextManager':
         return PoolItemContextManager(self)
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -20,7 +20,7 @@ class Pool:
     __slots__ = (
         'loop', '__max_size', '__items',
         '__constructor', '__created', '__lock',
-        '__constructor_args', '__close_lock'
+        '__constructor_args'
     )
 
     def __init__(self, constructor: ConstructorType, *args,
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -20,18 +20,19 @@ class Pool:
     __slots__ = (
         'loop', '__max_size', '__items',
         '__constructor', '__created', '__lock',
-        '__constructor_args'
+        '__constructor_args', '__close_lock'
     )
 
     def __init__(self, constructor: ConstructorType, *args,
                  max_size: int = None, loop: asyncio.AbstractEventLoop = None):
         self.loop = loop or asyncio.get_event_loop()
         self.__max_size = max_size
-        self.__items = asyncio.Queue(loop=self.loop)
+        self.__items = asyncio.Queue()
         self.__constructor = constructor
         self.__constructor_args = args or ()
         self.__created = 0
-        self.__lock = asyncio.Lock(loop=self.loop)
+        self.__lock = asyncio.Lock()
+        self.__close_lock = asyncio.Lock()
 
     def acquire(self) -> 'PoolItemContextManager':
         return PoolItemContextManager(self)
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -134,9 +134,18 @@ class RobustConnection(Connection):
         for number, channel in self._channels.items():
             try:
                 await channel.on_reconnect(self, number)
-            except CONNECTION_EXCEPTIONS:
+            except CONNECTION_EXCEPTIONS as e:
                 log.exception('Open channel failure')
-                await self.close()
+
+                if self.connection is not None:
+                    await asyncio.gather(
+                        self.connection.close(e), return_exceptions=True
+                    )
+                    self.connection = None
+
+                # The channel has not been opened successfully.
+                # Staying positive and just retry later.
+                await asyncio.sleep(self.reconnect_interval)
                 return
 
         self._reconnect_callbacks(self)
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -109,7 +109,9 @@ class RobustConnection(Connection):
             while True:
                 if self.connection is not None:
                     await asyncio.gather(
-                        self.connection.close(e), return_exceptions=True
+                        self.connection.close(
+                            RuntimeError("Reconnecting")
+                        ), return_exceptions=True
                     )
                     self.connection = None
 
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -44,6 +44,7 @@ class RobustConnection(Connection):
 
         self.__channels = set()
         self._reconnect_callbacks = CallbackCollection()
+        self._reconnect_lock = asyncio.Lock()
         self._closed = False
 
     @property
@@ -102,29 +103,34 @@ class RobustConnection(Connection):
 
     async def reconnect(self):
         if self.is_closed:
-            return
+            raise RuntimeError("{!r} connection closed".format(self))
 
-        try:
-            await super().connect()
+        async with self._reconnect_lock:
+            while True:
+                if self.connection is not None:
+                    await asyncio.gather(
+                        self.connection.close(e), return_exceptions=True
+                    )
+                    self.connection = None
 
-            for number, channel in self._channels.items():
-                await channel.on_reconnect(self, number)
+                try:
+                    await super().connect()
 
-            self._reconnect_callbacks(self)
+                    for number, channel in self._channels.items():
+                        await channel.on_reconnect(self, number)
 
-        except CONNECTION_EXCEPTIONS as e:
-            log.exception('Connection attempt error')
+                    self._reconnect_callbacks(self)
+                    return
+                except CONNECTION_EXCEPTIONS as e:
+                    log.exception('Connection attempt error')
 
-            if self.connection is not None:
-                await asyncio.gather(
-                    self.connection.close(e), return_exceptions=True
-                )
-                self.connection = None
+                    if self.connection is not None:
+                        await asyncio.gather(
+                            self.connection.close(e), return_exceptions=True
+                        )
+                        self.connection = None
 
-            self.loop.call_later(
-                self.reconnect_interval,
-                lambda: self.loop.create_task(self.reconnect())
-            )
+                await asyncio.sleep(self.reconnect_interval)
 
     def channel(self, channel_number: int = None,
                 publisher_confirms: bool = True,
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -74,7 +74,10 @@ class RobustConnection(Connection):
 
         super()._on_connection_close(connection, closing)
 
-        log.info("Connection to %s closed reconnecting...", self)
+        log.info(
+            "Connection to %s closed. Reconnecting after %r seconds.",
+            self, self.reconnect_interval
+        )
         self.loop.call_later(
             self.reconnect_interval,
             lambda: self.loop.create_task(self.reconnect())
@@ -105,7 +108,8 @@ class RobustConnection(Connection):
 
         if self.reconnecting:
             log.warning(
-                "Connect method called but connection %r is reconnecting now",
+                "Connect method called but connection %r is "
+                "reconnecting right now.",
                 self
             )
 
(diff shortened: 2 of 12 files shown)


aio_pika/robust_connection.py
```diff
@@ -134,9 +134,18 @@ def channel(self, channel_number: int = None,
         for number, channel in self._channels.items():
             try:
                 await channel.on_reconnect(self, number)
-            except CONNECTION_EXCEPTIONS:
+            except CONNECTION_EXCEPTIONS as e:
                 log.exception('Open channel failure')
-                await self.close()
+
+                if self.connection is not None:
+                    await asyncio.gather(
```
decaz: Why `gather` but not just `self.close(e)`?

aio_pika/robust_connection.py
```diff
@@ -134,9 +134,18 @@ def channel(self, channel_number: int = None,
...
+                        self.connection.close(e), return_exceptions=True
+                    )
+                    self.connection = None
+
+                # The channel has not been opened successfully.
+                # Staying positive and just retry later.
+                await asyncio.sleep(self.reconnect_interval)
```
decaz: I guess this has no effect.. Maybe we should get rid of `_on_reconnect` method and rewrite `reconnect` method as follows:
```python
    async def reconnect(self):
        if self.is_closed:
            return

        while True:
            try:
                await super().connect()
            except CONNECTION_EXCEPTIONS:
                log.exception('Connection attempt error')
                await asyncio.sleep(self.reconnect_interval)
            else:
                try:
                    for number, channel in self._channels.items():
                        await channel.on_reconnect(self, number)
                except CONNECTION_EXCEPTIONS as e:
                    log.exception('Open channel failure')
                    if self.connection is not None:
                        await self.close(e)
                        self.connection = None
                else:
                    break
        self._reconnect_callbacks(self)
```
What do you think?

aio_pika/robust_connection.py
mosquito: `self.close()` will set `self.__closed` flag. And connection couldn't be reopened after.

decaz: Ah, understand.

aio_pika/robust_connection.py
mosquito: I try this, it's broken. I don't want to refactor this part now, just fix production issue.

decaz: OK, but what this sleep gives here? I think the bug is `await self.close()` which prevents `reconnect` method to connect again (because of `self._closed` flag which you mentioned earlier).

decaz: If you meant waiting before reconnect attempt then there is already sleep at `_on_connection_close` callback..

mosquito: Yeah. You are right. After rewrite looks better.

aio_pika/pool.py
```diff
@@ -20,18 +20,19 @@ class Pool:
...
         self.loop = loop or asyncio.get_event_loop()
         self.__max_size = max_size
-        self.__items = asyncio.Queue(loop=self.loop)
+        self.__items = asyncio.Queue()
         self.__constructor = constructor
         self.__constructor_args = args or ()
         self.__created = 0
-        self.__lock = asyncio.Lock(loop=self.loop)
+        self.__lock = asyncio.Lock()
+        self.__close_lock = asyncio.Lock()
```
decaz: I guess this is for #269. Maybe remove within current PR?

aio_pika/robust_connection.py
```diff
@@ -106,15 +106,25 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
+                await channel.on_reconnect(self, number)
+
+            self._reconnect_callbacks(self)
+
+        except CONNECTION_EXCEPTIONS as e:
             log.exception('Connection attempt error')
 
+            if self.connection is not None:
+                await asyncio.gather(
+                    self.connection.close(e), return_exceptions=True
```
decaz: If you use `return_exceptions=True` to avoid exceptions raised maybe should use following:
```python
with suppress(CONNECTION_EXCEPTIONS):
    await self.connection.close(e)
```

aio_pika/pool.py
mosquito: Nope, commits from this PR will be removed from #269

decaz: I mean this attribute is not used.

aio_pika/robust_connection.py
```diff
@@ -102,19 +103,36 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
         if self.is_closed:
-            return
-
-        try:
-            await super().connect()
-        except CONNECTION_EXCEPTIONS:
-            log.exception('Connection attempt error')
+            raise RuntimeError("{!r} connection closed".format(self))
+
+        async with self._reconnect_lock:
```
decaz: Maybe we should check the lock before?
```python
if self._reconnect_lock.locked():
    log.warning('Already reconnecting')
    return

async with self._reconnect_lock:
    ...
```

aio_pika/robust_connection.py
mosquito: gather creates a separate task from coroutine. I think this better.

aio_pika/pool.py
mosquito: Yes. Done

mosquito:
@decaz @dizballanze all checks done, please review.

aio_pika/pool.py
decaz: `__close_lock` remained in `__slots__`.

aio_pika/robust_connection.py
```diff
@@ -64,6 +74,7 @@ def _on_connection_close(self, connection, closing, *args, **kwargs):
 
         super()._on_connection_close(connection, closing)
 
+        log.info("Connection to %s closed reconnecting...", self)
```
decaz: Little grammar fix: "Connection to %s **closed. R**econnecting..."

aio_pika/robust_connection.py
```diff
@@ -44,7 +44,13 @@ def __init__(self, url, loop=None, **kwargs):
 
         self.__channels = set()
         self._reconnect_callbacks = CallbackCollection()
+        self._connect_lock = asyncio.Lock()
         self._closed = False
+        self.connected = asyncio.Event()
```
decaz: What this event for?

mosquito: Event for external use. Useful for tests right now but I want to document it.

aio_pika/robust_connection.py
```diff
@@ -77,44 +91,61 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
-                if self.fail_fast:
-                    raise
-
-                log.warning(
-                    "First connection attempt failed "
-                    "and will be retried after %d seconds",
-                    self.reconnect_interval,
-                    exc_info=True,
-                )
+        if self.reconnecting:
```
decaz: Why not to return if already reconnecting?

mosquito: when the user wants to call reconnect explicitly it's just printed log warning.

aio_pika/robust_connection.py
```diff
@@ -77,44 +91,61 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
+    async def __cleanup_connection(self, exc):
+        await asyncio.gather(
+            self.connection.close(exc),
+            return_exceptions=True,
+        )
+        self.connection = None
+
     async def connect(self, timeout: TimeoutType = None, **kwargs):
+        if self.is_closed:
+            raise RuntimeError("{!r} connection closed".format(self))
```
decaz: This method is for connecting, maybe raising exception here is not needed?

decaz: Or it is should be at `reconnect` method.

aio_pika/robust_connection.py
```diff
@@ -64,6 +74,10 @@ def _on_connection_close(self, connection, closing, *args, **kwargs):
 
         super()._on_connection_close(connection, closing)
 
+        log.info(
+            "Connection to %s closed. Reconnecting after %r seconds.",
```
decaz: after **%d** seconds

aio_pika/robust_connection.py
mosquito: Nope. It's a case when the user wants to reopen closed connection.

aio_pika/robust_connection.py
mosquito: `%r` returns same result but without problem when non integer passed.

decaz: It's a little bit confusing that %s, %d and %r are mixed. I mean sometimes it is %d, but sometimes it is %r and so on..

aio_pika/robust_connection.py
decaz: Do you mean that if robust connection was closed then the user can't open it again?

mosquito: `self.is_closed` will be changed only after `self.close()` called. But never in reconnecting state.

aio_pika/robust_connection.py
mosquito: My rule is "Type `%r` when you log user-defined parameters"

aio_pika/robust_connection.py
decaz: I understand it. I mean following:
```python
connection = await aio_pika.connect_robust(...)
await connection.close()
await connection.connect()  # This would be illegal?
<RuntimeError>
```

mosquito: Yep illegal

aio_pika/robust_connection.py
decaz: I mean usage of "after %d seconds" below (line 137) but "after %r seconds" here for example. Anyway, let's merge it.

mosquito: Fine.

aio_pika/robust_connection.py
```diff
@@ -133,10 +133,8 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
                     await self.__cleanup_connection(e)
 
                     log.warning(
-                        "Connection attempt to \"%s\" failed "
-                        "and will be retried after %d seconds",
```
decaz: Why did you remove this line? `_on_connection_close` doesn't log it while reconnecting..

mosquito: `self._cleanup_connection` will log same log.

decaz: You mean here? https://github.com/mosquito/aio-pika/blob/1065083434c4ef3b45c653e30bc78e8637b8dad1/aio_pika/robust_connection.py#L77-L80
But it won't log while reconnecting due to https://github.com/mosquito/aio-pika/blob/1065083434c4ef3b45c653e30bc78e8637b8dad1/aio_pika/robust_connection.py#L63-L65

mosquito: Do you want to swap it?

mosquito: `_on_connection_close` will be called when every connection closed. In [this case](https://github.com/mosquito/aio-pika/blob/1065083434c4ef3b45c653e30bc78e8637b8dad1/aio_pika/robust_connection.py#L63-L65) connection is establishing and this condition needed only for `self.__cleanup_connection`.

decaz: I thought you want to log about reconnection interval here https://github.com/mosquito/aio-pika/blob/1065083434c4ef3b45c653e30bc78e8637b8dad1/aio_pika/robust_connection.py#L77-L80
OK, but now after removal of lines about it I don't see how `self._cleanup_connection` would log it, - there is no any logging in it.

mosquito: What do you propose? A simple way is copying log from [here](https://github.com/mosquito/aio-pika/blob/1065083434c4ef3b45c653e30bc78e8637b8dad1/aio_pika/robust_connection.py#L77-L80) to `connect` function I guess.

decaz: I suggest to save _"and will be retried after %d seconds"_ part of log within `connect` method. I don't see how it will be logged twice.

aio_pika/robust_connection.py
```diff
@@ -77,44 +91,61 @@ def add_reconnect_callback(self, callback: Callable[[], None]):
...
+
+                    await self.__cleanup_connection(e)
+
+                    log.warning(
+                        "Connection attempt to \"%s\" failed. "
+                        "Reconnecting after %r seconds.",
+                        self, self.reconnect_interval, exc_info=True,
+                    )
+                except asyncio.CancelledError as e:
+                    await self.__cleanup_connection(e)
```
decaz: @mosquito shouldn't we break the loop here?

PR #268, opened 2019-11-15
Connection name query parameter

mosquito:
Add name query parameter for human readable connection name


mosquito:
part of aiormq

Issue #281, opened 2019-12-23
How to get exchange by name, not declare?


heckad:
Hello everybody. I have a big app wich declare all neede queues and exchanges, and I have the microservice witch should use it. I don't want to keep consistent configuration in this service, because I need methods for getting these resources. Are there any methods that by name, allow you to get a queue or exchange?

mosquito:
Just call `.declare()` with `passive=True` flag. It's should work for `Queues` and `Exchanges`

heckad:
Is it update the configuration or continue to keep old?

mosquito:
It's just get current configuration or fails otherwise (e.g when entity doesn't exists).

heckad:
As I understand it, if you add `passive=True`, then the rest of the parameters are ignored (except name). Maybe you add methods `get_queue(name:str)` and `get_exchange(name:str)` for more readable code.

mosquito:
`.declare()` methods it’s an implementation of internal AMQP RPC frames. It’s still close to AMQP protocol and still readable. Of course I could add `.get()` method as a shortcut, but you can add your own inherited classes.

heckad:
It would be great if such methods will be available as I proposed in the previous comment.

ebrahimiali:
> Just call `.declare()` with `passive=True` flag. It's should work for `Queues` and `Exchanges`

Is there a similar way when using _**STOMP**_ protocol? It seems user needs configure permission when subscribe using stomp. It seems strange that it considers some parameters like `durable:true` or `autoDelete:true` but not `passive:true`

PR #282, opened 2019-12-24
Add get methods for exchange and queue

heckad:
https://github.com/mosquito/aio-pika/issues/281

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -225,6 +225,7 @@ class Channel:
 
     async def get_exchange(self, name: str) -> Queue:
         """
+        It's a shortcut for ``.declare_exchange(..., passive=True)``.
 
         :param name: queue name
         :return: :class:`aio_pika.queue.Queue` instance
@@ -271,6 +272,7 @@ class Channel:
 
     async def get_queue(self, name: str) -> Queue:
         """
+        It's a shortcut for ``.declare_queue(..., passive=True)``.
 
         :param name: queue name
         :return: :class:`aio_pika.queue.Queue` instance
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -223,13 +223,21 @@ class Channel:
 
         return exchange
 
+    async def get_exchange(self, name: str) -> Queue:
+        """
+
+        :param name: queue name
+        :return: :class:`aio_pika.queue.Queue` instance
+        """
+
+        return await self.declare_exchange(name=name, passive=True)
+
     async def declare_queue(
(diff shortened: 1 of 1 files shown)


heckad:
Where to add tests?

aio_pika/channel.py
```diff
@@ -223,13 +223,21 @@ def _on_return(self, message: aiormq.types.DeliveredMessage):
 
         return exchange
 
+    async def get_exchange(self, name: str) -> Queue:
+        """
+
```
mosquito:
```
It's a shortcut for ``.declare_exchange(..., passive=True)``.
```

heckad: Yes

mosquito: Correct. Please add it as a docstring.

heckad: I didn’t understand. Do you want copy code from 233 line and put it to 228?

mosquito: Could you please add docstring for new methods.

heckad:
@mosquito, do I need to add something else?

aio_pika/channel.py
```diff
@@ -223,13 +223,22 @@ def _on_return(self, message: aiormq.types.DeliveredMessage):
 
         return exchange
 
+    async def get_exchange(self, name: str) -> Queue:
+        """
+        It's a shortcut for ``.declare_exchange(..., passive=True)``.
+
+        :param name: queue name
```
decaz: exchange name

aio_pika/channel.py
```diff
@@ -223,13 +223,22 @@ def _on_return(self, message: aiormq.types.DeliveredMessage):
...
+        :return: :class:`aio_pika.queue.Queue` instance
```
decaz: aio_pika.exchange.Exchange

heckad:
@mosquito ?

heckad:
@mosquito ?

heckad:
@mosquito, please. It's the very simple pull request. What's wrong with them?

mosquito:
I investigated this case, and it seems that it does not work as we assume here.

if there is no entity in RabbitMQ, then get will create it. Unfortunately, this cannot be influenced.

mosquito:
Just add test which gets an undeclared queue, and after that try to declare it with different params.
User assumes another behaviour.

mosquito:
@decaz @heckad what we will do?

heckad:
I asked the question how to correct to do it on https://groups.google.com/forum/#!topic/pika-python/z-vwXsm9YuQ

PR #290, opened 2020-01-23
fix type annotation and documentation for Connection.add_close_callback

### LINKED ISSUE #289: incorrect type annotation on add_close_callback's callback parameter
The signature of `Connection.add_close_callback` is currently as follows:

`def add_close_callback(self, callback: Callable[[], None]):`

According to the docstring, the `callback` will be passed an `asyncio.Future` as the first argument. The current signature implies that it will be called with no arguments.

I think the correct annotation is:
`def add_close_callback(self, callback: Callable[[asyncio.Future], None]):`
or if you would rather not specify the arguments:
`def add_close_callback(self, callback: Callable[[...], None]):`

Happy to open a PR if this sounds ok.

brianmaissy:
Fixes #289

### DIFF
--- a/aio_pika/connection.py
+++ b/aio_pika/connection.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 from functools import partial
-from typing import Callable, Type, TypeVar, Optional
+from typing import Type, TypeVar
 
 from yarl import URL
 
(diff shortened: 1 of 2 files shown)


brianmaissy:
It appears that this just missed a release. Any idea when the next one will be?

PR #303, opened 2020-03-03
Fix call close on closed channels

heckad:
If write something like

```
rabbitmq_connection: aio_pika.Connection = await aio_pika.connect_robust(
    f"amqp://{rabbitmq_option('username')}:{rabbitmq_option('password')}@{rabbitmq_option('host')}/"
)

async with rabbitmq_connection:
    async with rabbitmq_connection.channel() as channel:
        pass
```

You got warning `Channel already closed`

To solve this problem, I added a channel closedness check.

### DIFF
--- a/aio_pika/connection.py
+++ b/aio_pika/connection.py
@@ -205,8 +205,9 @@ class Connection:
         return self
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
-        for channel in tuple(self._channels.values()):
-            await channel.close()
+        for channel in tuple(self._channels.values()):  # _channel can change size
+            if not channel.is_closed:
+                await channel.close()
 
         await self.close()
 
--- a/aio_pika/connection.py
+++ b/aio_pika/connection.py
@@ -205,7 +205,7 @@ class Connection:
         return self
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
-        for channel in tuple(self._channels.values()):  # _channel can change size
+        for channel in tuple(self._channels.values()):  # can change size
             if not channel.is_closed:
                 await channel.close()
 


heckad:
@mosquito, you can review.

aio_pika/connection.py
```diff
@@ -205,8 +205,9 @@ def __del__(self):
         return self
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
-        for channel in tuple(self._channels.values()):
-            await channel.close()
+        for channel in tuple(self._channels.values()):  # can change size
+            if not channel.is_closed:
+                await channel.close()
```
mosquito: Hmm... that looks pretty simple but when you have to do this is the user closes channel manually. So, it's reasonable, but I imagine this simpler than the current patch.

```python
while (self._channels):
    channel = self._channels.pop()
    if channel.is_closed:
        continue
    await channel.close()
```

It's might be a guarantee of avoiding another edge case, a user opening a new channel while `` await channel.close()`` performing.

Another way is adding a lock or condition and trigger it while close is starting. But the first will be better I guess.

heckad: I just noticed that we have a potential memory leak as it was in https://github.com/mosquito/aio-pika/pull/285. What if the user opens the channels and closes them himself? Do we clear `_channels` from closed channels?

mosquito: Seems to be yes.

heckad: Then why I got `Channel already closed` if we delete it from `_channels`?

heckad: If add `print(sys.getrefcount(channel))` onto 211 line into connection then will be print `6`? Why so many references?

heckad: If disable `gc` then objects don't delete. It's a memory leak.

mosquito (approved):

PR #306, opened 2020-03-09
Fix #302

### LINKED ISSUE #302: message.reject called inside ProcessContext.__exit__ fails when channel is closed
I'm using ProcessContext in my consumer implementation and get error:

```
AttributeError
'NoneType' object has no attribute 'write'
```
in https://github.com/mosquito/aiormq/blob/bedf52cee682777435c71486e1614636aedc9330/aiormq/channel.py#L411

This exception is raised during handling of [ChannelInvalidStateError](https://github.com/mosquito/aiormq/blob/bedf52cee682777435c71486e1614636aedc9330/aiormq/channel.py#L89) raised inside the body of ProcessContext context manager.

ProcessContext tries to [reject](
https://github.com/mosquito/aio-pika/blob/3915d2fd8e9c377197b5613bd4f5aeb6035a2826/aio_pika/message.py#L673) the message but fails to do that because the channel is closed.

DXist:
This PR fixes issue #302 happening during message reject on closed channel.

Now `ChannelInvalidStateError` is propagated through process context manager and user is expected to handle this exception. Usually it means just to ignore this error since current message should be redelivered and reprocessed again.


aio_pika/message.py
```diff
@@ -645,12 +649,22 @@ def __init__(self, message: IncomingMessage, *,
...
-
-                await self.message.reject(requeue=False)
+                if not self.message.channel.is_closed:
+                    log.info(
+                        "Message %r was redelivered and will be rejected",
+                        self.message
+                    )
+                    await self.message.reject(requeue=False)
+                else:
+                    log.info(
```
mosquito: It's should be `log.warning`, so it looks important.

aio_pika/message.py
```diff
@@ -665,12 +679,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                        self.message
+                    )
+
+                    self.message.reject(requeue=False)
+                else:
+                    log.info(
+                        "Message %r was redelivered and reject is not sent "
+                        "since channel is closed", self.message
+                    )
             else:
```
mosquito: Maybe add return here and remove `else`

aio_pika/message.py
```diff
@@ -645,12 +649,22 @@ def __init__(self, message: IncomingMessage, *,
...
+                        "Message %r was redelivered and will be rejected",
+                        self.message
+                    )
+                    await self.message.reject(requeue=False)
+                else:
+                    log.info(
+                        "Message %r was redelivered and reject is not sent "
+                        "since channel is closed", self.message
+                    )
             else:
```
mosquito: Maybe add return here and remove `else`.

aio_pika/message.py
DXist: @mosquito I've applied the changes you've suggested.

PR #307, opened 2020-03-12
WIP: Fix circular dependency in callbacks

heckad:


heckad:
@mosquito What to do [with]( https://deepsource.io/gh/mosquito/aio-pika/run/d103ecab-f758-4f41-8101-949d054d9b40/python/PYL-W0613). Now all callbacks should get sender.

aio_pika/channel.py
```diff
@@ -50,17 +50,17 @@ def __init__(self, connection, channel_number: int = None,
         self.loop = connection.loop
 
         self._connection = connection
-        self._done_callbacks = CallbackCollection()
-        self._return_callbacks = CallbackCollection()
-        self._channel = None  # type: aiormq.Channel
+        self._done_callbacks = CallbackCollection(self)
+        self._return_callbacks = CallbackCollection(self)
+        self._channel: Optional[aiormq.Channel] = None
```
mosquito: This syntax not supported for Python 3.5:

```
$ python3.5
Python 3.5.4 (v3.5.4:3f56838976, Aug  7 2017, 12:56:33)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> f: bool = True
  File "<stdin>", line 1
    f: bool = True
     ^
SyntaxError: invalid syntax
>>>
```

heckad: If revert then pycharm highlight that the wrong type. If write `type: Optional[aiormq.Channel]` then DeepSource [will swear](https://deepsource.io/gh/mosquito/aio-pika/run/42d515a2-ebae-43eb-8f28-6cc8c6973a25/python/PYL-C0412). What to do?

heckad:
I can't use WeakSet because callback `on_message_returned` in RPC should bound with rpc and this bound don't have any external references. For this case, in other libraries add a flag (for example `weak=False`) which says that need to store the callback in the set instead of the weak set? What do you think about this? Have you looked at blinker lib?

heckad:
@mosquito do you know why my test was skipped?

mosquito:
Please rebase.

heckad:
Wait. You add this branch in merge commit 25355b1f51802e1b8296b8f55b18cbcd5afdc1d5.

decaz:
@mosquito I guess these changes were accidentally merged into the master branch and therefore included in the latest 6.6.1 release.

PR #309, opened 2020-03-21
Remove not needed creation of new dict

heckad:


Gera3dartist:
won't this change,  create a reference to `self.headers._headers` instead of copying it?

heckad:
No, look at the Message constructor. 
```
        self.headers_raw = format_headers(headers)
        self._headers = HeaderProxy(self.headers_raw)
```
And format_headers
```
def format_headers(d: Dict[str, Any]) -> Dict[str, bytes]:
    ret = {}

    if not d:
        return ret

    for key, value in d.items():
        ret[key] = header_converter(value)
    return ret
```
We creates new  dict again!

heckad:
@mosquito?

mosquito:
Do not understand why is safe. The copied message can modify headers and original headers will be modified too.

heckad:
No, you do copy twice: in __copy__ and in format_headers.

mosquito:
Oh, my fail.

PR #311, opened 2020-03-28
Fix RobustChannel not reconnecting on duplicate named queue/exchanges

### LINKED ISSUE #229: RobustChannel will not reconnect if the queue/exchange is declared twice with the same name.
If you declare exchange for the second time (with the same parameters), it overrides the previous exchange in `self._exchanges[name]` reference, which causes "on_reconnect" not to pass new channel to declared exchanges/queues, resulting in `RuntimeError: <Channel: "1"> closed` errors.

I would make a pull request but don't know what's the correct solution, so I will propose few below:

I will state the examples on **exchanges**, but everything also applies to **queues**

**Option 1 - Fetch exchange from already declared dict**

`robust_channel.py`
```
async def declare_exchange(self, name: str,
                           type: ExchangeType=ExchangeType.DIRECT,
                           durable: bool=None, auto_delete: bool=False,
                           internal: bool=False, passive: bool=False,
                           arguments: dict=None, timeout: int=None,
                           robust: bool=True) -> Exchange:

+   if name in self._exchanges:
+       return self._exchanges[name]

    exchange = await super().declare_exchange(
        name=name, type=type, durable=durable, auto_delete=auto_delete,
        internal=internal, passive=passive, arguments=arguments,
        timeout=timeout,
    )

    if not internal and robust:
        self._exchanges[name] = exchange

    return exchange
```

Cons: 
- You return object which is used elsewhere
- If you declare exchange with different parameters it will silently pass and return current instance, instead of raising exception.

**Option 2 - variance on Option 1**
Actually trying to declare the exchange and discarting fetched object afterward 
`robust_channel.py`
```
async def declare_exchange(self, name: str,
                           type: ExchangeType=ExchangeType.DIRECT,
                           durable: bool=None, auto_delete: bool=False,
                           internal: bool=False, passive: bool=False,
                           arguments: dict=None, timeout: int=None,
                           robust: bool=True) -> Exchange:

    exchange = await super().declare_exchange(
        name=name, type=type, durable=durable, auto_delete=auto_delete,
        internal=internal, passive=passive, arguments=arguments,
        timeout=timeout,
    )

+   if name in self._exchanges:
+       return self._exchanges[name]

    if not internal and robust:
        self._exchanges[name] = exchange

    return exchange
```

Cons: 
- You return object which is used elsewhere
- Discarting awaited object.
- Confusing code without explanation/comments

**Option 3 - Compare all exchange connection parameters**
`robust_channel.py`
```
async def declare_exchange(self, name: str,
                           type: ExchangeType=ExchangeType.DIRECT,
                           durable: bool=None, auto_delete: bool=False,
                           internal: bool=False, passive: bool=False,
                           arguments: dict=None, timeout: int=None,
                           robust: bool=True) -> Exchange:

+   if name in self._exchanges:
+       exchange = self._exchanges[name]
+       if exchange.__type != type or exchange.durable is not durable or ... # rest
+           raise RuntimeException("can't declare exchange... different arguments") 
+       return self._exchanges[name]

    exchange = await super().declare_exchange(
        name=name, type=type, durable=durable, auto_delete=auto_delete,
        internal=internal, passive=passive, arguments=arguments,
        timeout=timeout,
    )

    if not internal and robust:
        self._exchanges[name] = exchange

    return exchange
```

**Option 4 - Store all the created exchanges**

`robust_channel.py`
```
async def declare_exchange(self, name: str,
                           type: ExchangeType=ExchangeType.DIRECT,
                           durable: bool=None, auto_delete: bool=False,
                           internal: bool=False, passive: bool=False,
                           arguments: dict=None, timeout: int=None,
                           robust: bool=True) -> Exchange:

    exchange = await super().declare_exchange(
        name=name, type=type, durable=durable, auto_delete=auto_delete,
        internal=internal, passive=passive, arguments=arguments,
        timeout=timeout,
    )

    if not internal and robust:
+       self._exchanges[(name, id(exchange))] = exchange

    return exchange
```

Cons:
- `exchange_delete` can't be done just by name - since you can have multiple instance with same name 
- OR delete all the references to the exchange with the name, in which case either make self._exchanges[name] = list() for easier delete lookup, but harder to cycle through (2 levels of nesting)

I'm inclined to Option 2 or Option 3, but I'd like to know other opinions or Options.

aviramha:
Fixes #229 
I wasn't sure if to add tests as my approach is pretty simple...

### DIFF
--- a/aio_pika/robust_channel.py
+++ b/aio_pika/robust_channel.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from typing import Union
 
 import aiormq
@@ -42,8 +43,8 @@ class RobustChannel(Channel):
             on_return_raises=on_return_raises,
         )
 
-        self._exchanges = dict()
-        self._queues = dict()
+        self._exchanges = defaultdict(set)
+        self._queues = defaultdict(set)
         self._prefetch_count = 0
         self._prefetch_size = 0
         self._global_ = False
@@ -59,11 +60,13 @@ class RobustChannel(Channel):
             global_=self._global_
         )
 
-        for exchange in self._exchanges.values():
-            await exchange.restore(self)
+        for exchanges in self._exchanges.values():
+            for exchange in exchanges:
+                await exchange.restore(self)
 
-        for queue in self._queues.values():
-            await queue.restore(self)
+        for queues in self._queues.values():
+            for queue in queues:
+                await queue.restore(self)
 
     def _on_channel_close(self, exc: Exception):
         if exc:
@@ -108,7 +111,7 @@ class RobustChannel(Channel):
         )
 
         if not internal and robust:
-            self._exchanges[name] = exchange
+            self._exchanges[name].add(exchange)
 
         return exchange
 
@@ -138,7 +141,7 @@ class RobustChannel(Channel):
         )
 
         if robust:
-            self._queues[name] = queue
+            self._queues[name].add(queue)
 
         return queue
 
@@ -151,6 +154,7 @@ class RobustChannel(Channel):
         )
 
         self._queues.pop(queue_name, None)
+
         return result
 
 
--- a/tests/test_amqp_robust.py
+++ b/tests/test_amqp_robust.py
@@ -321,6 +321,54 @@ class TestCase(AMQPTestCase):
             pass
         await queue.unbind(exchange, routing_key)
 
+    async def test_robust_duplicate_queue(self):
+        queue_name = "test"
+        channel1 = await self.create_channel()
+        channel2 = await self.create_channel()
+
+        shared = []
+        queue1 = await channel1.declare_queue(queue_name)
+        queue2 = await channel1.declare_queue(queue_name)
+
+        async def reader(queue):
+            nonlocal shared
+            async with queue.iterator() as q:
+                async for message in q:
+                    shared.append(message)
+                    await message.ack()
+
+        reader_task1 = self.loop.create_task(reader(queue1))
+        reader_task2 = self.loop.create_task(reader(queue2))
+        self.addCleanup(reader_task1.cancel)
+        self.addCleanup(reader_task2.cancel)
+
+        for _ in range(5):
+            await channel2.default_exchange.publish(
+                Message(b''), queue_name,
+            )
+
+        logging.info("Disconnect all clients")
+        await self.proxy.disconnect()
+
+        logging.info("Waiting for reconnect")
+        await asyncio.sleep(5)
+
+        logging.info("Waiting connections")
+        await asyncio.wait([
+            channel1._connection.ready(),
+            channel2._connection.ready()
+        ])
+
+        for _ in range(5):
+            await channel2.default_exchange.publish(
+                Message(b''), queue_name,
+            )
+
+        while len(shared) < 10:
+            await asyncio.sleep(0.1)
+
+        self.assertEqual(len(shared), 10)
+
 
 class TestCaseNoRobust(AMQPTestCase):
     async def create_connection(self, cleanup=True) -> Connection:


aio_pika/robust_channel.py
```diff
@@ -42,8 +43,8 @@ def __init__(self, connection, channel_number: int = None,
             on_return_raises=on_return_raises,
         )
 
-        self._exchanges = dict()
-        self._queues = dict()
+        self._exchanges = []
```
mosquito, line 46: `defaultdict(set)` will looks better I guess.

aviramha: But why should it be a dict?

aio_pika/robust_channel.py
```diff
@@ -121,7 +122,8 @@ def _on_channel_close(self, exc: Exception):
             if_unused=if_unused, nowait=nowait
         )
 
-        self._exchanges.pop(exchange_name, None)
+        self._exchanges[:] = filterfalse(lambda x: x.name == exchange_name,
```
mosquito, line 125: This change unneeded when `self._exchanges` is an instance of dict (defaultdict)

aviramha: I think it's less nice then just a list, but OK. updated the PR

aio_pika/robust_channel.py
```diff
@@ -108,7 +111,7 @@ def _on_channel_close(self, exc: Exception):
         )
 
         if not internal and robust:
-            self._exchanges[name] = exchange
+            self._exchanges.append(exchange)
```
mosquito, line 114: `self._exchanges[name].add(exchange)`

aio_pika/robust_channel.py
```diff
@@ -121,7 +124,7 @@ def _on_channel_close(self, exc: Exception):
             if_unused=if_unused, nowait=nowait
         )
 
-        self._exchanges.pop(exchange_name, None)
+        self._exchanges.pop(exchange_name)
```
mosquito, line 127: revert it please

aio_pika/robust_channel.py
```diff
@@ -150,7 +153,8 @@ def _on_channel_close(self, exc: Exception):
             if_unused=if_unused, if_empty=if_empty, nowait=nowait
         )
 
-        self._queues.pop(queue_name, None)
+        self._queues.pop(queue_name)
```
mosquito, line 156: revert it please

aviramha:
Done

mosquito:
Looks simple and nice, @decaz @dizballanze, please review.

dizballanze:
Hi, @aviramha. Maybe you could write a test for this?

aviramha:
Master is broken or something? The tests fail on chunks of code that existed beforehand.

dizballanze:
@mosquito

aviramha:
@mosquito Any ETA on the next version release? :)

mosquito:
[6.6.1](https://pypi.org/project/aio-pika/6.6.1/) has been released

aviramha:
For some reason I can't find on master the changes I've done and it's not in the changel log also. Was the PR reverted?

mosquito:
See #323 I'm applied your original commit so you are co-author

aviramha:
Ah I thought the fix is in master already.

Sinkler:
Hello!

After this changing we detected a memory leak in our project. It looks like there is a problem in `defaultdict(set)`. We replaced it by `defaultdict(WeakSet)` and now it's ok:

```python
class FixRobustChannel(RobustChannel):
    def __init__(
        self,
        connection,
        channel_number: int = None,
        publisher_confirms: bool = True,
        on_return_raises=False,
    ):
        super().__init__(connection, channel_number, publisher_confirms, on_return_raises)
        self._exchanges = defaultdict(WeakSet)
        self._queues = defaultdict(WeakSet)


class FixRobustConnection(RobustConnection):
    CHANNEL_CLASS = FixRobustChannel


await aio_pika.connect_robust("...", connection_class=FixRobustConnection)
```

PR #321, opened 2020-05-17
Add generics to Pool and PoolItemContextManager

taybin:
This helps mypy determine the return type when using a pool like:
`async with self.connection_pool.acquire() as connection:`
or
`async with self.channel_pool.acquire() as channel:`

Without this, mypy was telling me to set type for the `connection` and `channel` variables.

### DIFF
--- a/aio_pika/pool.py
+++ b/aio_pika/pool.py
@@ -2,7 +2,7 @@ import abc
 import asyncio
 import logging
 from typing import (
-    Any, AsyncContextManager, Awaitable, Callable, Coroutine, TypeVar, Union,
+    Any, AsyncContextManager, Awaitable, Callable, Coroutine, Generic, TypeVar, Union,
 )
 
 from aiormq.tools import awaitable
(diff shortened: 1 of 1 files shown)


mosquito:
@decaz any thoughts? For me, it's looking very nice.

aio_pika/pool.py
```diff
@@ -30,7 +31,7 @@ class PoolInvalidStateError(RuntimeError):
     pass
 
 
-class Pool:
+class Pool(Generic[T]):
```
decaz, line 34: Is it necessary here? I don't see any usage of `Pool[T]`.

taybin: It's used by the user of the aio-pika library.

```
class Messenger:
    connection_pool: Pool[aio_pika.Connection]
    channel_pool: Pool[aio_pika.Channel]
```

taybin:
@decaz I removed the unused `ItemType` declaration.

mosquito:
released with [6.6.1](https://pypi.org/project/aio-pika/6.6.1/)

PR #322, opened 2020-05-20
Fix Docs

nhumrich:


mosquito:
released with [6.6.1](https://pypi.org/project/aio-pika/6.6.1/)

gyermolenko:
@mosquito Hi
(before this PR) I noticed that ReadTheDocs tags end with 4.8.0
image
Can you tell why this is the last tag and what "latest" is pointing to? 
Thanks.

mosquito:
Oh my bad. Now I set the 6.6.1 tag.

Issue #325, opened 2020-06-02
Timeout not work with queue iterator


heckad:
```
        async with queue.iterator(timeout=10) as queue_iter:
            async for message in queue_iter:
                async with message.process():
                    print(message.body)
```
This code works endlessly

romuald:
@heckad thanks for the fix!

Any idea when the next release will be?

PR #326, opened 2020-06-02
Fix timeout onto read messages

heckad:
https://github.com/mosquito/aio-pika/issues/325


heckad:
Wait, I found a more critical error. If the connection breaks, the consumer will not know about it. Do you have any idea how to forward an exception to a consumer?

heckad:
Seems class Queue don't have a close method which should close all Queue iterators.

Issue #314, opened 2020-04-20
RobustConnection Reconnecting even if manually closed


cyberlis:
After updating to version 6.6.0 I get errors, when I try to close connection
```python
import asyncio
import logging

from aio_pika.robust_connection import connect_robust


logging.basicConfig(level=logging.DEBUG)


async def main():
    connection = await connect_robust()
    await connection.close()


if __name__ == "__main__":
    asyncio.run(main())
```

Output
```
DEBUG:asyncio:Using selector: KqueueSelector
DEBUG:aiormq.connection:Can not read bytes from server:
Traceback (most recent call last):
  File "/Users/d.lisovik/Library/Caches/pypoetry/virtualenvs/dispipe-2Qe6ZZMa-py3.7/lib/python3.7/site-packages/aiormq/connection.py", line 377, in __reader
    weight, channel, frame = await self.__receive_frame()
  File "/Users/d.lisovik/Library/Caches/pypoetry/virtualenvs/dispipe-2Qe6ZZMa-py3.7/lib/python3.7/site-packages/aiormq/connection.py", line 329, in __receive_frame
    frame_header = await self.reader.readexactly(1)
  File "/Users/d.lisovik/.pyenv/versions/3.7.2/lib/python3.7/asyncio/streams.py", line 677, in readexactly
    raise IncompleteReadError(incomplete, n)
asyncio.streams.IncompleteReadError: 0 bytes read on a total of 1 expected bytes
DEBUG:aio_pika.connection:Closing AMQP connection None
INFO:aio_pika.robust_connection:Connection to amqp://guest:******@localhost:5672// closed. Reconnecting after 5 seconds.
```

I found one problem with connection closing.
Output
```
DEBUG:aio_pika.connection:Closing AMQP connection None
```

This problem in `_make_connection`
```python
async def _make_connection(self, **kwargs) -> aiormq.Connection:
    connection = await aiormq.connect(self.url, **kwargs)
    connection.closing.add_done_callback(
        partial(self._on_connection_close, self.connection)
    )
    return connection
```
Specifically in
```python
partial(self._on_connection_close, self.connection)
```
When we call partial, `self.connection` is `None` at this particular moment
To fix this we can pass `connection` instead of `self.connection`

But this doesn't fix Reconnecting and IncompleteReadError problem

PR #329, opened 2020-06-17
Conditionally reconnect RobustConnection

soulmerge:
RobustConnection was reconnecting whenever the connection was lost,
even if the connection had been closed on pupose by invoking
`connection.close()`. The reconnect code now tests if a reconnect
attempt should be made at all by consulting `self._closed`.

### DIFF
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -75,6 +75,9 @@ class RobustConnection(Connection):
 
         super()._on_connection_close(connection, closing)
 
+        if self._closed:
+            return
+
         log.info(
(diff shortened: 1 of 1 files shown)


soulmerge:
Related: #314

soulmerge:
I just noticed that this was already fixed at one point, but the fix was lost during a larger code cleanup:

https://github.com/mosquito/aio-pika/commit/b446be3a582dd93cc7ed29e6e3f14e95103d2ece#diff-d45310da4c963431973c61cb287a348a

soldag:
Any idea when will this change be released?

PR #351, opened 2020-10-09
Add `ensure` parameter to `get_exchange`/`get_queue`

wyfo:
When True, this parameter doesn't change the methods behavior; when False, methods just return Exchange/Queue instance without declaring it, saving a round-trip.


mosquito:
Related to #353

mosquito (approved):

PR #363, opened 2020-11-16
drop 3.5 support

mosquito:


decaz (changes requested):
@mosquito I guess that https://github.com/mosquito/aio-pika/blob/5af608d05b05d65632652f0632d00af4729b265b/aio_pika/exceptions.py#L14-L23 should be rewritten as the following:
```python
CONNECTION_EXCEPTIONS = (
    RuntimeError,
    ConnectionError,
    OSError,
    AMQPError,
    pamqp.exceptions.PAMQPException,
)
```

aio_pika/exceptions.py
```diff
@@ -11,9 +11,7 @@
 )
 
 
-PAMQP_EXCEPTIONS = (pamqp.exceptions.PAMQPException,) + tuple(
-    pamqp.specification.ERRORS.values(),
-)
+PAMQP_EXCEPTIONS = (pamqp.exceptions.PAMQPException,)
```
decaz, line 14: Can we get rid of this tuple? It was created to collect all pamqp exceptions because it wasn't trivial until now.

aio_pika/exceptions.py
```diff
@@ -11,9 +11,7 @@
...
 
 CONNECTION_EXCEPTIONS = (
```
decaz, line 16: I think `ConnectionError` can be removed because it inherits `OSError` which is already exists here.

aio_pika/exceptions.py
mosquito: done

aio_pika/channel.py
```diff
@@ -6,6 +6,7 @@
 
 import aiormq
 import aiormq.types
+from pamqp import commands
```
decaz, line 9: In other files you are making `import pamqp.commands`, - I think it worth to do the same here so that it'll be the same everywhere.

decaz:
@mosquito from pamqp's [changelog](https://github.com/gmr/pamqp/blob/master/docs/changelog.rst#300a0-2019-12-16): `Changed pamqp.decode.timestamp to return a datetime.datetime instance instead of time.struct_time.`.

So I guess all helpers (`encode_timestamp` and `decode_timestamp`) can be removed, isn't it?

PR #364, opened 2020-11-19
Add example of reopening channels that have been closed

bluesliverx:
This bit me recently when channels were closed unexpectedly due to long running message retrieval on an API where client connections were closed suddenly. This apparently put the channels in a bad state (`ChannelInvalidStateError("writer is None")` to be exact), and therefore I had to check channels when retrieving them from the pool.

### DIFF
--- a/docs/source/examples/pooling-recycled.py
+++ b/docs/source/examples/pooling-recycled.py
@@ -1,0 +1,78 @@
+from collections import defaultdict
+
+import asyncio
+import aio_pika
+from aio_pika.pool import Pool
+
+
+class NonRestoringRobustChannel(aio_pika.RobustChannel):
+    """
+    Custom robust channel that when reopened, does not restore any
+    queues or exchanges.
+    """
+    async def reopen(self) -> None:
+        # Clear out exchanges and queues when reopened
+        self._exchanges = defaultdict(set)
+        self._queues = defaultdict(set)
+        await super().reopen()
+
+
+class NonRestoringRobustConnection(aio_pika.RobustConnection):
+    """
+    Robust connection that uses a custom channel class
+    """
+    CHANNEL_CLASS = NonRestoringRobustChannel
+
+
+async def main():
+    loop = asyncio.get_event_loop()
+
+    async def get_connection():
+        return await aio_pika.connect_robust(
+            "amqp://guest:guest@localhost/",
+            # Use the connection class that does not restore connections
+            connection_class=NonRestoringRobustConnection,
+        )
+
+    connection_pool = Pool(get_connection, max_size=2, loop=loop)
+
+    async def get_channel() -> aio_pika.Channel:
+        async with connection_pool.acquire() as connection:
+            return await connection.channel()
+
+    channel_pool = Pool(get_channel, max_size=10, loop=loop)
+    queue_name = "pool_queue"
+
+    async def consume():
+        async with channel_pool.acquire() as channel:  # type: aio_pika.Channel
+            await channel.set_qos(10)
+
+            queue = await channel.declare_queue(
+                queue_name, durable=False, auto_delete=False
+            )
+
+            async with queue.iterator() as queue_iter:
+                async for message in queue_iter:
+                    print(message)
+                    await message.ack()
+
+    async def publish():
+        async with channel_pool.acquire() as channel:  # type: aio_pika.Channel
+            # Reopen channels that have been closed previously
+            if channel.is_closed:
+                await channel.reopen()
+            await channel.default_exchange.publish(
+                aio_pika.Message(("Channel: %r" % channel).encode()),
+                queue_name,
+            )
+
+    async with connection_pool, channel_pool:
+        task = loop.create_task(consume())
+        await asyncio.wait([publish() for _ in range(10000)])
+        await task
+
+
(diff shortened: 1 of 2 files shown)


docs/source/examples/pooling.py
```diff
@@ -33,6 +33,9 @@ async def consume():
 
     async def publish():
         async with channel_pool.acquire() as channel:  # type: aio_pika.Channel
+            # (Optional) Reopen channels that have been closed previously
+            if channel.is_closed:
+                channel.reopen()
```
Sinkler, line 38:
```suggestion
                await channel.reopen()
```

bluesliverx:
Thanks @Sinkler for catching that. I found that I actually needed some more infrastructure around this to work properly since channels were attempting to reinitialize queues and exchanges on reopen. I've changed my commit to make sure things look good.

docs/source/examples/pooling-recycled.py
```diff
@@ -0,0 +1,78 @@
...
+
+class NonRestoringRobustChannel(aio_pika.RobustChannel):
+    """
+    Custom robust channel that when reopened, does not restore any
+    queues or exchanges.
+    """
+    async def reopen(self) -> None:
+        # Clear out exchanges and queues when reopened
+        self._exchanges = defaultdict(set)
+        self._queues = defaultdict(set)
```
Sinkler, line 16: By the way, we had some memory leaks with `set` so we replaced it by `WeakSet` in the `__Init__` method and now it's ok 🙃

bluesliverx:
@mosquito the py35/py36-uvloop tox tests are failing, but I'm not sure what to do about those. They appear to be unrelated to my PR.

docs/source/quick-start.rst
```diff
@@ -51,3 +51,9 @@ Connection pooling
 
 .. literalinclude:: examples/pooling.py
    :language: python
+
+Connection pooling with recycled channels
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
mosquito, line 57: Please add some context here. Why it's useful? When pool might be used without this example? Why `Connect` class is not a replacement for `NonRestoringRobustConnection`? etc.

This will be as useful as the code itself.

docs/source/examples/pooling-recycled.py
bluesliverx: Thanks for the info! I will make those same changes...

docs/source/quick-start.rst
bluesliverx: Added

docs/source/examples/pooling-recycled.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class NonRestoringRobustChannel(aio_pika.RobustChannel):
+    """
+    Custom robust channel that when reopened, does not restore any
+    queues or exchanges.
+    """
+    async def reopen(self) -> None:
+        # Clear out exchanges and queues when reopened
+        self._exchanges = defaultdict(set)
```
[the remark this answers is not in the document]
bluesliverx: Good call, changed.

bluesliverx:
@mosquito I have updated the PR if you could take a look again. Thanks!

mosquito (approved):

Sinkler (approved):

bluesliverx:
Ran across this again, any chance of getting it merged @mosquito?

mosquito:
@bluesliverx ouch, I should be write this comment earlier, but doesn't sorry. I completely do not understand why I will want to apply this example in real project when to be honest. When you want do not reconnecting when the underlay-connection faills you shouldn't use `connect_robust`, just use `connect`.

bluesliverx:
@mosquito, I *do* want the robust connection since I want it to re-established. However, I do *not* want the queues and exchanges to be declared again since some of them are exclusive and cannot be recreated by a new connection. I believe rabbit threw errors in that case.

As for why I need the `channel.is_closed` check, I'm not sure why the robust channel doesn't handle this internally, but I did find it was necessary when using pools. In certain cases such as clients closing connections mid-use, the channel was not robust enough to reconnect on its own and needed the check to reopen.

IsaacFung:
Any chance of following up on this @mosquito? Thank you!

mosquito:
@bluesliverx If some kind of check is required for reopening, it's a bug, you need to learn how to reproduce it, write a test and fix it.

As for exclusive consumers, Rabbit itself should release it after two failed heartbeats. It doesn't do this because for some reason it didn't receive a signal about closing the TCP socket from the client.

dsconnol-impactai:
I haven't run into this, but I came across this issue while trying to ensure robustness while migrating to the aio-pika. 

As far as I can tell, this is specifically an issue for publishing, because the RobustExchange doesn't implement it's own `publish` method, so it never calls `await channel.ready()` like the other methods on robust classes - leading to this potential failure mode. 

I'm happy to put together a PR at some point to fix that - but before doing so I wanted to check whether that's an intentional design decision (i.e. I'm misunderstanding the library) or an accidental oversight that deserves a fix.

mosquito:
@dsconnol-impactai you should try, if tests will pass after this change, it must be ok. I written all design decisions in tests, I hope)))

PR #369, opened 2020-12-30
Update rpc.py

Sobolev5:
Python3.9 rpc.py fix

### DIFF
--- a/aio_pika/patterns/rpc.py
+++ b/aio_pika/patterns/rpc.py
@@ -153,7 +153,7 @@ class RPC(Base):
             self.on_message_returned, weak=False,
         )
 
-    def on_close(self, exc=None):
+    def on_close(self, exc=None, _=None):
         log.debug("Closing RPC futures because %r", exc)
         for future in self.futures.values():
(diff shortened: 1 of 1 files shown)


aio_pika/patterns/rpc.py
```diff
@@ -153,7 +153,7 @@ async def initialize(self, auto_delete=True, durable=False, **kwargs):
             self.on_message_returned, weak=False,
         )
 
-    def on_close(self, exc=None):
+    def on_close(self, exc=None, _=None):
```
mosquito, line 156: Could you please add some details why python 3.9 adding an argument here?

ivunchata: I believe this is not due to python 3.9, rather it is a regression from `1625f37e4056c355e9b20962f39f8ba6e66e953a`.
I have tried with python 3.6 onward and the attached Dockerfile reproduces the issue that is solved by this PR.
[Dockerfile.zip](https://github.com/mosquito/aio-pika/files/7198735/Dockerfile.zip)

Issue #371, opened 2021-01-14
Add reject messages which were ack in a transaction if the transaction fails.


heckad:
Now If I write 
```
    async with message.process(requeue=True), channel.transaction():
        message.ack()
        
        raise Exception
```
I got MessageProcessError because ack method set `__processed` in true. Also, I can't call reject manually after the transaction for the same reason. Propose to store ack message in transaction `WeakSet` and if we do rollback then set  `__processed` into false and add flag `reject_ack_message_on_rollback` or something similar name which will reject all backed messages.

mosquito:
just add `async with message.process(requeue=True, ignore_processed=True), ...:`

heckad:
It doesn't do auto message reject.

heckad:
@mosquito does channel can have more than one transaction?

PR #372, opened 2021-01-14
Basic impl of restore processed flag after rollback

heckad:
https://github.com/mosquito/aio-pika/issues/371


mosquito:
You must add `contextvars` as an extra dependency for python 3.5 and 3.6.

PR #389, opened 2021-05-10
Prevent 'Set size changed during iteration'

### LINKED ISSUE #388: Error: "Set size changed during iteration"
We experienced this in production a few times (screenshot from Sentry):

image

My guess is that this is due to the async nature of aio-pika; that while iterating through the set, the await call inside the iteration body causes a possibility for that coroutine to yield, and another to run. If `declare_exchange` is called before the set iteration is complete, I think this could happen...

Edit: Our version of aio-pika is 6.7.1

torarvid:
Fixes #388 (possibly?)

Honestly, I'm not sure how to test this, but it feels like the way to
(safely) iterate through a set in an async context, is to iterate
through a copy of the set (or use a lock to that modification of the
set is blocked until iteration is complete).

### DIFF
--- a/aio_pika/robust_channel.py
+++ b/aio_pika/robust_channel.py
@@ -67,7 +67,7 @@ class RobustChannel(Channel):
         await self.default_exchange.restore(self)
 
         for exchanges in self._exchanges.values():
-            for exchange in exchanges:
+            for exchange in exchanges.copy():
                 await exchange.restore(self)
 
         for queues in self._queues.values():


aio_pika/robust_channel.py
```diff
@@ -67,7 +67,7 @@ async def restore(self) -> None:
         await self.default_exchange.restore(self)
 
         for exchanges in self._exchanges.values():
-            for exchange in exchanges:
+            for exchange in exchanges.copy():
```
mosquito, line 70: This narrows down the scope of refactoring a bit in the future. In my opinion, it would be enough to do `tuple(exchanges)` here. `tuple()` requires something iterable as input and can not modifying.

mosquito (changes requested):

mosquito:
This fix will work well, however it doesn't seem to solve the original problem. The fact that the set changes at the time of recovery is very strange.

The resulting custom code declares or deletes exchange during connection recovery.

mosquito:
@torarvid sorry for delay, it was a vacation, and there was no time for Open Source.

PR #393, opened 2021-06-02
Make Python version compatible with PEP 345

albireox:
This just changes the `python_requires` in `setup.py` to use the version definition supported by PEP 345.

See [here](https://github.com/python-poetry/poetry/issues/4095#issuecomment-851065187) for how this has become an issue.


finswimmer (approved):

albireox:
It seem the PY35 tests fail because `multidict` dropped support for Python 3.5 [last year](https://github.com/aio-libs/multidict/blob/master/setup.py). Is there something I should do there?

aviramha:
@mosquito can you merge this as well please?

PR #395, opened 2021-06-13
Fix connection passed to Connection._on_connection_close

zaquest:
Seems like the `connection` argument is used solely for logging, still it's weird to see `Closing AMQP connection None` all the time.


tilsche:
I saw this as well. It is pretty distracting during debugging. Would be great if this could be merged.

PR #415, opened 2021-09-25
reduce verbosity of reconnect warnings

tilsche:
Replace the full traceback with a simple inline string of the exception of the failed reconnect attempt.
The full traceback requires many lines and appear frequently in logs (default every 5 s) at high visibility (warning).
At the same time, the traceback offers very little useful extra information in addition to the exception string.


volbil:
I was about to create issue regarding traceback spam but saw this PR. Thank you very much 👍

PR #417, opened 2021-10-04
Support dict inside headers

levrik:
After switching from `pika` to `Kombu` and then to `aio-pika` I've discovered that `aio-pika` doesn't properly supported nested dicts inside headers like the other 2 libraries.
Supporting it was quite easy.


levrik:
@mosquito Could this be merged? We started to look for alternatives to Kombu once again and noticed that my PR still wasn't merged which is blocking us from using this library.

mosquito:
@levrik I'm afraid that you will have to add a test for this, that the received message contains the correct header payload

dizballanze (approved):

decaz (approved):

mosquito (approved):

mosquito:
released in 8.2.5

PR #436, opened 2022-01-24
Feature/modern aiormq

mosquito:
A really great PR that reworks class inheritance by adding abstract classes, and checks everything with `mypy`.

**Will be released as 7.0.x**


aio_pika/abc.py
```diff
@@ -0,0 +1,658 @@
...
+class DeclarationResult(NamedTuple):
+    message_count: int
+    consumer_count: int
+
+
+class AbstractTransaction:
+    state: TransactionStates
+
+    @abstractproperty
+    def channel(self) -> "AbstractChannel":
```
Pavkazzz, line 66: `-> aiormq.abc.AbstractChannel:`

mosquito: Nope, Transaction returns aio-pike's Channel instance. aiormq's channel is AMQP transport object and not exposes from aio-pika's classes.

aio_pika/abc.py
```diff
@@ -0,0 +1,658 @@
...
+
+class DeclarationResult(NamedTuple):
+    message_count: int
+    consumer_count: int
+
+
+class AbstractTransaction:
+    state: TransactionStates
+
+    @abstractproperty
```
dizballanze, line 65: `abstractproperty` is deprecated, use `abstractmethod` and `property` instead

aio_pika/queue.py
```diff
@@ -225,24 +226,28 @@ async def consume(
...
+                loop=self.loop,
+            ),
+            exclusive=exclusive,
+            no_ack=no_ack,
+            arguments=arguments,
+            consumer_tag=consumer_tag,
+            timeout=timeout,
+        )
+
+        assert consume_result.consumer_tag is not None, "Consumer tag is None"
```
dizballanze, line 244: Why do you use assert outside of tests?

aio_pika/queue.py
```diff
@@ -381,64 +384,95 @@ async def main():
...
-                return
+        log.debug("Queue iterator %r closed", self)
+
+        def queue_tail(
+            channel: aiormq.abc.AbstractChannel,
+        ) -> Generator[Any, AbstractIncomingMessage, None]:
+            while not channel.is_closed:
+                try:
+                    yield self._queue.get_nowait()
+                except asyncio.QueueEmpty:
```
dizballanze, line 415: `contextlib.suppress`?

aio_pika/channel.py
```diff
@@ -56,158 +55,180 @@ def __init__(
...
 
-        self._connection = connection
-        self._done_callbacks = CallbackCollection(self)
-        self._return_callbacks = CallbackCollection(self)
-        self._channel = None  # type: Optional[aiormq.Channel]
+        self._channel: aiormq.abc.AbstractChannel
         self._channel_number = channel_number
+
+        self.connection = connection
+        self.close_callbacks = CallbackCollection(self)
```
Pavkazzz, line 62: Why not private?

aio_pika/channel.py
```diff
@@ -56,158 +55,180 @@ def __init__(
...
             publisher_confirms=self._publisher_confirms,
             on_return_raises=self._on_return_raises,
             channel_number=self._channel_number,
+            timeout=timeout,
         )
 
     async def initialize(self, timeout: TimeoutType = None) -> None:
-        if self._channel is not None:
-            raise RuntimeError("Can't initialize channel")
+        if self.is_opened:
```
Pavkazzz, line 177: Maybe add some verbolize in is_close and use only that? 
Always checks for `is_opened` and `_is_closed_by_user` looks verbose and very easy to forget one of this

aio_pika/channel.py
```diff
@@ -56,158 +55,180 @@ def __init__(
...
 
-    def __await__(self):
+    def __await__(self) -> Generator[Any, Any, "AbstractChannel"]:
         yield from self.initialize().__await__()
         return self
 
-    async def __aenter__(self):
-        if not self._channel:
+    async def __aenter__(self) -> "AbstractChannel":
+        if not hasattr(self, "_channel"):
```
Pavkazzz, line 134: Why not simple None, instead of checking attr?
Some GC feature?

aio_pika/message.py
```diff
@@ -1,213 +1,221 @@
...
 
 
 @singledispatch
-def header_converter(value: Any) -> bytes:
-    return json.dumps(
-        value, separators=(",", ":"), ensure_ascii=False, default=repr,
-    ).encode()
+def header_converter(value: Any) -> FieldValue:
+    return bytearray(
+        json.dumps(
```
Pavkazzz, line 185: Maybe give options to choose json encoder for client?

aio_pika/patterns/master.py
```diff
@@ -173,8 +187,9 @@ async def create_worker(
         return Worker(queue, consumer_tag, self.loop)
 
     async def create_task(
-        self, channel_name: str, kwargs=None, **message_kwargs
-    ):
+        self, channel_name: str,
+        kwargs: Dict[str, Any] = None, **message_kwargs: Any
```
Pavkazzz, line 191: `kwargs: Optional[Dict[str, Any]] = None`

setup.py
```diff
@@ -29,21 +29,20 @@
         "Operating System :: Microsoft",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
Pavkazzz, line 35: 3.10 maybe too?

Pavkazzz (changes requested):

aio_pika/channel.py
mosquito: I want to declare management methods (e.g. `add_close_callback`) deprecated since 8.0

aio_pika/channel.py
mosquito: It's semantically different cases. `is_opened` means the channel is ready for interaction. `is_closed` means the channel has been closed.

aio_pika/channel.py
mosquito: Do not want to make it `Optional` and check that every time.

aio_pika/message.py
mosquito: I don't think that transferring a small number of bytes to the header will not cause a performance problem (or gain).

aio_pika/channel.py
Pavkazzz: But you make it not existing at all and must check's all the time :)

dizballanze (approved):

aio_pika/channel.py
mosquito: C'est La Vie

aio_pika/abc.py
```diff
@@ -0,0 +1,672 @@
...
+from .pool import PoolInstance
+from .tools import CallbackCollection
+
+
+TimeoutType = Optional[Union[int, float]]
+
+NoneType = type(None)
+DateType = Union[int, datetime, float, timedelta, None]
+ExchangeParamType = Union["AbstractExchange", str]
+ConsumerTag = str
```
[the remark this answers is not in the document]
mosquito: Yes, it's actually the `TypeAlias` but `typing.TypeAlias` is available only since python 3.10.

aio_pika/abc.py
```diff
@@ -0,0 +1,672 @@
...
+from pamqp.common import Arguments
+from yarl import URL
+
+from .pool import PoolInstance
+from .tools import CallbackCollection
+
+
+TimeoutType = Optional[Union[int, float]]
+
+NoneType = type(None)
```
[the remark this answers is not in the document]
mosquito: yep

aio_pika/types.py
```diff
@@ -1,9 +1,11 @@
...
+from .abc import *  # noqa
 
 
-Sender = Any
-ReturnCallbackType = Callable[[Sender, aiormq.types.DeliveredMessage], Any]
-CloseCallbackType = Callable[[Sender, Optional[BaseException]], None]
-TimeoutType = Union[int, float]
+warnings.warn(
+    "aio_pika.types was deprecated and will be removed in "
+    "one of next major releases. Use aio_pika.abc instead.",
```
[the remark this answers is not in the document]
mosquito: `types` is available in 6.x and removing it might breaks user code.

aio_pika/connection.py
```diff
@@ -96,19 +101,37 @@ async def main():
...
+            self.closing.set_exception(exc)
+            return
+
+        self.closing.set_result(closing.result())
+
+    async def _make_connection(
+        self, *, timeout: TimeoutType = None, **kwargs: Any
+    ) -> aiormq.abc.AbstractConnection:
+        connection: aiormq.abc.AbstractConnection = await asyncio.wait_for(
+            aiormq.connect(self.url, **kwargs), timeout=timeout,
```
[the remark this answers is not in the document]
mosquito: aiormq is low level communication driver and the `asyncio.wait_for` should works fine in any cases

aio_pika/connection.py
```diff
@@ -1,114 +1,125 @@
...
+from typing import Any, Callable, Dict, Optional, Tuple, Type, TypeVar, Union
+from warnings import warn
 
 import aiormq
+from aiormq.abc import ExceptionType
 from aiormq.tools import censor_url
 from yarl import URL
 
+# For migration from 6.x to 7.x
+from .abc import ConnectionType  # noqa
```
mosquito, line 14: I didn't find any other way to tell the user that the import from this file is deprecated, so I removed it from `__all__` and now only the user linter (or IDE) will tell him about it. But I tried to ensure that the user code did not break during migration.

aio_pika/connection.py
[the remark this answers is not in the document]
mosquito: Let me explain what I mean. The `timeout` parameter in the RPC-related methods is necessary since it uses a slightly more complex cancellation mechanism than sending `CancelledError` to the coroutine.

Therefore, this parameter is present there. In all other cases, async io.wait_for will work.

aio_pika/connection.py
```diff
@@ -192,32 +219,33 @@ async def main(loop):
...
-        if any((self.is_closed, self.loop.is_closed(), not self.connection)):
+    def __del__(self) -> None:
+        if (
+            self.is_closed or
+            self.loop.is_closed() or
+            not hasattr(self, "connection")
+        ):
             return
 
         asyncio.shield(self.close())
```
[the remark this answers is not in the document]
mosquito: you can't make __del__ is awaitable. Shield creates a non-cancellable task and returns a separate Future that joined to the task.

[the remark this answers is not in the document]
decaz: I understand that but I'm used to await tasks to prevent them from being orphaned and errors like `Task was destroyed but it is pending!`.

aio_pika/connection.py
decaz: As I said earlier, - 7.0 is a new major version so that is normal if it will break something.

aio_pika/connection.py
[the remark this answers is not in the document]
decaz: Okay.

decaz (changes requested):
@mosquito look at all comments, please.

aio_pika/abc.py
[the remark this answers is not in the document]
mosquito: should pass type to `singledispatch`. When you trying to pass `None` directly you have an exception
```
TypeError: Invalid first argument to `register()`: None. Use either `@register(some_class)` or plain `@register` on an annotated function.
```

aio_pika/abc.py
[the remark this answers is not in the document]
mosquito: For example, in AMQP 0.9 specification might be clarified how the consumer-tag string is formed, then it will no longer be just a string. This is an alias for the future so that the user uses an explicitly specified type instead of a string in this particular case.

aio_pika/abc.py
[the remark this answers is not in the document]
decaz: Ok then!

aio_pika/connection.py
mosquito: Yes but really sad when you upgrade a library and have to fix all your code and run all tests (etc.) to get it done. 

In my opinion, people love upgrades but really hate just technical changes which borns incompatible changes.

Moving the type definitions is not impacted to the memory consumption performance or whatever, and for me, it's a question of upgrade adoption. I have no plans for support 6.X branch right now cause I want to provide the simplest upgrade way.

This release contains changes in library core and promises minimal public API changes.

Probably you are right and I have to write these promises explicitly in the CHANGELOG.

setup.py
```diff
@@ -29,21 +29,21 @@
...
+        "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: Implementation :: PyPy",
         "Programming Language :: Python :: Implementation :: CPython",
     ],
     packages=find_packages(exclude=["tests"]),
     package_data={"aio_pika": ["py.typed"]},
-    install_requires=["aiormq>=3.2.3,<4", "yarl"],
-    python_requires=">=3.5, <4",
+    install_requires=["aiormq~=6.1.1", "yarl"],
+    python_requires=">3.6.*, <4",
```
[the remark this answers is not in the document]
mosquito: It's excluding pre-releases when I correctly understand it.

aio_pika/connection.py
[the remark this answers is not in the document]
mosquito: When you have not forgotten a non-closed object, it will not be called

aio_pika/abc.py
```diff
@@ -0,0 +1,671 @@
...
+    reply_to: Optional[str]
+    expiration: Optional[DateType]
+    message_id: Optional[str]
+    timestamp: Optional[datetime]
+    type: Optional[str]
+    user_id: Optional[str]
+    app_id: Optional[str]
+
+    @property
+    def headers(self) -> HeadersType:
```
Alviner, line 127: Missed `abstractmethod`?

aio_pika/message.py
```diff
@@ -1,213 +1,217 @@
...
 @singledispatch
-def encode_expiration(value) -> Optional[str]:
+def encode_expiration(value: Any) -> Optional[str]:
     raise ValueError("Invalid timestamp type: %r" % type(value), value)
 
 
 @encode_expiration.register(datetime)
-def _(value):
+def encode_expiration_datetime(value: datetime) -> str:
     now = datetime.now()
```
Alviner, line 39: What if `offset-naive datetimes` will be passed as expiration?

aio_pika/patterns/rpc.py
```diff
@@ -3,36 +3,44 @@
...
-R = TypeVar("R")
-P = TypeVar("P")
-CallbackType = Callable[[P], R]
+T = TypeVar("T")
+CallbackType = Callable[..., T]
 
 
-class RPCMessageTypes(Enum):
+class RPCMessageType(str, Enum):
     error = "error"
```
Alviner, line 34: Maybe change to project enum style?

aio_pika/abc.py
```diff
@@ -0,0 +1,672 @@
...
+    @abstractmethod
+    async def consume(self) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def __aiter__(self) -> "AbstractQueueIterator":
+        raise NotImplementedError
+
+    @abstractmethod
+    def __aenter__(self) -> Awaitable["AbstractQueueIterator"]:
```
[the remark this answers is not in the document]
mosquito: This returning awaitable object which broad than a coroutine.

aio_pika/channel.py
```diff
@@ -56,175 +55,184 @@ def __init__(
...
-        if not self._channel:
+    async def __aenter__(self) -> "AbstractChannel":
+        if not self.is_opened:
             await self.initialize()
-
         return self
 
-    async def __aexit__(self, exc_type, exc_val, exc_tb):
-        await self.close()
+    def __aexit__(
```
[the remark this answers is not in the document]
mosquito: This returning awaitable object which broad than a coroutine.

aio_pika/tools.py
```diff
@@ -1,19 +1,20 @@
 import asyncio
 import logging
-from collections.abc import Set
 from functools import wraps
 from itertools import chain
 from threading import Lock
```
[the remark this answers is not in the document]
mosquito: Threading lock guarantees the execution in one thread. You can execute a callback in another thread or run an event-loop in another thread than the main.

In a normal situation, this lock is always released and does not lock anything, but ensures correct internal object states.

aio_pika/tools.py
```diff
@@ -92,39 +111,39 @@ def add(self, callback: Callable, weak=True):
...
+    def discard(self, callback: Callable[..., Any]) -> None:
         if self.is_frozen:
             raise RuntimeError("Collection frozen")
 
         with self.__lock:
             try:
-                self.__callbacks.remove(callback)
+                self.__callbacks.remove(callback)    # type: ignore
             except KeyError:
                 self.__weak_callbacks.remove(callback)
```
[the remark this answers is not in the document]
mosquito: Since the lock is acquired, we can check conditions.

aio_pika/tools.py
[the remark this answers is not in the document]
decaz: @mosquito I don't quite understand why `asyncio.Lock` can't be used instead? If user needs to run something in another thread then it can be done with `asyncio.to_thread`.

PR #439, opened 2022-02-20
copy entities when restoring

mosquito:


decaz (approved):

aio_pika/robust_exchange.py
```diff
@@ -56,7 +56,7 @@ async def restore(self, channel: AbstractRobustChannel) -> None:
 
         await self.declare()
 
-        for exchange, kwargs in self._bindings.items():
+        for exchange, kwargs in tuple(self._bindings.items()):
```
alvassin, line 59: Why do you need explicit tuple here?

alvassin (approved):

PR #452, opened 2022-03-31
fixes #447

### LINKED ISSUE #447: Crash when the logger is set to debug mode
First of all, thanks for the library!

I was playing with it and trying to work with properly deleting my queues / channels / connections when cancelling some tasks. I enabled the debug logger, and noticed that the event loop would stop, due to an exception. I traced it to those three locations, within `queue.py`:

- The queue is being printed [here](https://github.com/mosquito/aio-pika/blob/28b445240043a7e11f13421093f97ca144cf9a4b/aio_pika/queue.py#L411) using `%r` formatting in a debug message
- `repr` is defined [here](https://github.com/mosquito/aio-pika/blob/28b445240043a7e11f13421093f97ca144cf9a4b/aio_pika/queue.py#L429), and basically prints the class name, the queue name and the consumer tag
- However the consumer tag gets [deleted a few lines above](https://github.com/mosquito/aio-pika/blob/28b445240043a7e11f13421093f97ca144cf9a4b/aio_pika/queue.py#L406)

This means that at that point, the class has no more `_consumer_tag` member, and all attempts to print it will fail. I can see two ways to fix it, but I'd like your opinion:
- Modify `repr` to check for the presence / absence of `consumer_tag` using `hasattr`.
- Modify the way `_consumer_tag` is handled, and set it to None instead of deleting it. This means we need to change all checks.

mosquito:


dizballanze (approved):

Alviner (approved):

decaz (changes requested):
Maybe it worth just to add private flag `is_cancelled` instead of this `hasattr` & `getattr` error prone magic combination?

mosquito:
@decaz Unfortunately, this will have no effect in this particular case. Since a lazy logger can still have time to break

PR #456, opened 2022-04-20
Feature/rpc extending

mosquito:


alvassin (approved):

aio_pika/patterns/base.py
```diff
@@ -34,7 +34,21 @@ class Base:
     CONTENT_TYPE = "application/python-pickle"
 
     def serialize(self, data: Any) -> bytes:
+        """ Serialize data to the bytes.
+        Uses `pickle` by default.
+        You should overlap this method when you want to change serializer
+
+        :param data: Data which will be serialized
+        :returns: bytes
```
decaz, line 42: Is specification of returned types is necessary within doctstring? There is already typing exists.

aio_pika/patterns/master.py
```diff
@@ -99,7 +99,7 @@ def __init__(
         :param channel: Initialized instance of :class:`aio_pika.Channel`
         """
         self.channel: AbstractChannel = channel
-        self.loop: asyncio.AbstractEventLoop = self.channel.loop
```
decaz, line 102: Why not to continue to use existing loop value?

aio_pika/patterns/rpc.py
```diff
@@ -288,38 +286,34 @@ async def on_call_message(
...
-
-    def serialize_exception(self, exception: Exception) -> bytes:
-        """ Serialize python exception to bytes
-
-        :param exception: :class:`Exception`
-        :return: bytes
-        """
-        return pickle.dumps(exception)
+    def serialize_exception(self, exception: Exception) -> Any:
+        """ Serialize python exception to bytes """
```
decaz, line 290: Due to the name of the method and due to the docstring says it serializes exception to bytes it's expected that it wouldn't return exception as is but there is no serialization and it returns `Any` type..

decaz (changes requested):

aio_pika/patterns/rpc.py
mosquito: Cause for the JsonRPC it's a dictionary. This method serialises an exception, not returning an exception.

aio_pika/patterns/master.py
mosquito: Yeah it's will conflicts with future releases

aio_pika/patterns/base.py
mosquito: this docstrings just moved.

aio_pika/patterns/rpc.py
decaz: Then "to bytes" needs to be removed from docstring, I guess.

mosquito:
@decaz done

decaz (approved):

Issue #481, opened 2022-07-18
Feature request: providing amqp "connection.update_secret" implementation for refreshing tokens 


deschmih:
Hey, 
while connecting to rabbitmq and using jwt in an oauth2 manner, a token refresh before expiring is necessary. Currently I can do this only in the way closing the connection and open again. 
Since AMQP 0.9.1 https://www.rabbitmq.com/amqp-0-9-1-reference.html there is another option to do this. Connection.update-secret is the preferred way to do this. 
Update-secret is implemented in pamqp library. It would be nice to have this feature available in aio-pika.connection as a separate method.

mosquito:
@deschmih feel free to  join the [review](https://github.com/mosquito/aiormq/pull/145).

deschmih:
You have done a good really job. Many Thanks for this. 
But I think you should also bump version in setup.py to aiormq~=6.4.0
https://github.com/mosquito/aio-pika/blob/8c1515127d37df5997d869c5129d119d5f65e12a/setup.py#L41

deschmih:
@mosquito To get a tagged version would be the icing on the cake. ;-)
Many thanks.

PR #483, opened 2022-07-22
bump version to aiormq~=6.4.0  in setup.py

deschmih:
because of comment https://github.com/mosquito/aio-pika/issues/481#issuecomment-1189961369 in issue


mosquito:
This is not enough.  I'm waiting for the weekend to work on this properly, for example for the RobustConnection you have to update the secret so that it works when reconnection.

mosquito:
Duplicate of #484

PR #486, opened 2022-07-25
Generated anonymous queue name may conflict (fix #485)

### LINKED ISSUE #485: Generated anonymous queue name may conflict if random seed is fixed (with connect_robust)
## Issue

Because `RobustQueue._get_random_queue_name()` uses python's global `random generator`, it always produce the **same anonymous queue name** if user has a fixed random seed. Which is certainly not desired.

And a library should not restrict users to use the global random generator, so I think this is a bug.

### How I encounter this issue
I declared some callback queue to be **anonymous** and **exclusive**. If the same queue name is already declared, I get this error message.

```
aiormq.exceptions.ChannelLockedResource: RESOURCE_LOCKED - cannot obtain exclusive access to locked queue 'amq_0xe5712ff7671b26d9377140c8cfbffcbb' in vhost '/'. 
It could be originally declared on another connection or the exclusive property value does not match that of the original declaration.
```

## Reproduce

*You will always get same queue name*

``` python 
import asyncio
import random

import aio_pika


async def main():
    random.seed(0)
    conn = await aio_pika.connect_robust('amqp://localhost')
    channel = await conn.channel()
    queue = await channel.declare_queue()
    print(queue.name)


if __name__ == '__main__':
    asyncio.run(main())

```

## Fix

Use a standalone random generator should fix this problem.

``` python
# aio_pika/robust_queue.py
class RobustQueue(Queue, AbstractRobustQueue):
    __slots__ = ("_consumers", "_bindings")

    _consumers: Dict[ConsumerTag, Dict[str, Any]]
    _bindings: Dict[Tuple[Union[AbstractExchange, str], str], Dict[str, Any]]

    _rnd_gen: random.Random = random.Random()

    @classmethod
    def _get_random_queue_name(cls) -> str:
        rnd = cls._rnd_gen.getrandbits(128)
        return "amq_%s" % hex(rnd).lower()
```

manfred-exz:
Fix #485

### DIFF
--- a/aio_pika/robust_queue.py
+++ b/aio_pika/robust_queue.py
@@ -1,4 +1,4 @@
-from random import getrandbits
+from random import getrandbits, Random
 from typing import Any, Callable, Dict, Optional, Tuple, Union
 
 import aiormq
@@ -23,9 +23,11 @@ class RobustQueue(Queue, AbstractRobustQueue):
     _consumers: Dict[ConsumerTag, Dict[str, Any]]
     _bindings: Dict[Tuple[Union[AbstractExchange, str], str], Dict[str, Any]]
 
-    @staticmethod
-    def _get_random_queue_name() -> str:
-        rnd = getrandbits(128)
+    _rnd_gen: Random = Random()
+
+    @classmethod
+    def _get_random_queue_name(cls) -> str:
+        rnd = cls._rnd_gen.getrandbits(128)
         return "amq_%s" % hex(rnd).lower()
 
     def __init__(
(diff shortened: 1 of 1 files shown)


aio_pika/robust_queue.py
```diff
@@ -1,4 +1,4 @@
-from random import getrandbits
+from random import getrandbits, Random
```
mosquito, line 1: remove useless import `getrandbits`

mosquito (commented):
remove useless import

decaz (approved):

mosquito:
@manfred-exz released in `aio-pika==8.1.1`

PR #491, opened 2022-09-09
Fix memory leaks on channel close

### LINKED ISSUE #490: Memory leak on robust channel opening-closing
Our application constantly opens and closes channels using robust connection.
We detected memory leak, it can be reproduced using following snippet:

```import argparse
import asyncio

import aio_pika

CHANNELS_COUNT = 1000


async def main(amqp_uri):
    connection = await aio_pika.connect_robust(amqp_uri)
    print(len(connection.transport.connection.closing._callbacks))
    for _ in range(CHANNELS_COUNT):
        async with connection.channel():
            await asyncio.sleep(0)
    print(len(connection.transport.connection.closing._callbacks))


if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument("--amqp_uri")
    args = parser.parse_args()
    asyncio.run(main(args.amqp_uri))
```

Not an expert in aio_pika, but seems leak can be fixed in abc.UnderlayConnection: callbacks should be removed on channel closing with .remove_done_callback, smth like:
```
    async def close(self, exc: Optional[ExceptionType] = None) -> Any:
        if self.close_callback.finished.is_set():
            return

        self.channel.closing.remove_done_callback(self.close_callback)
        self.channel.connection.closing.remove_done_callback(self.close_callback)

        result: Any = await self.channel.close(exc)
        await self.close_callback.wait()
        return result
```

mosquito:
Fixes #490 

`leaks_test.py`
```python
import asyncio
import logging

import aiomisc
from aiomisc.service.tracer import MemoryTracer
from aio_pika import connect_robust


async def main():
    logging.basicConfig(level=logging.DEBUG)

    counter = 0
    while True:
        connect = await connect_robust()
        channel = await connect.channel()
        await asyncio.sleep(0.001)
        await channel.close()
        await connect.close()
        counter += 1

        if not counter % 100:
            logging.info("%d Connections done", counter)


aiomisc.run(main(), MemoryTracer(), log_format="rich")
```

### DIFF
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -459,9 +459,18 @@ class UnderlayChannel(NamedTuple):
     async def close(self, exc: Optional[ExceptionType] = None) -> Any:
         if self.close_callback.finished.is_set():
             return
-        await self.channel.close(exc)
+
+        # close callbacks must be fired when closing
+        # and should be deleted later for prevent memory leaks
+        result: Any = await self.channel.close(exc)
         await self.close_callback.wait()
-        return
+
+        self.channel.closing.remove_done_callback(self.close_callback)
+        self.channel.connection.closing.remove_done_callback(
+            self.close_callback
+        )
+
+        return result
 
 
 class AbstractChannel(PoolInstance, ABC):
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -461,7 +461,7 @@ class UnderlayChannel(NamedTuple):
             return
 
         # close callbacks must be fired when closing
-        # and should be deleted later for prevent memory leaks
+        # and should be deleted later to prevent memory leaks
         await self.channel.close(exc)
         await self.close_callback.wait()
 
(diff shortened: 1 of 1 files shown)


Alviner (dismissed):

dizballanze (approved):

aio_pika/abc.py
```diff
@@ -459,9 +459,16 @@ async def create(
...
+        # close callbacks must be fired when closing
+        # and should be deleted later for prevent memory leaks
         await self.channel.close(exc)
         await self.close_callback.wait()
-        return
+
+        self.channel.closing.remove_done_callback(self.close_callback)
+        self.channel.connection.closing.remove_done_callback(
+            self.close_callback
+        )
```
decaz, line 471: Isn't it necessary to do the same in the `UnderlayConnection.close` method?

decaz (changes requested):

aio_pika/abc.py
mosquito: Not reproduced.

decaz (approved):

PR #492, opened 2022-09-20
Prevent "Task exception was never retrieved" on timeout

torarvid:
Sometimes we have an issue with RabbitMQ in production, and we see a lot of these types of errors:

image

This appears to be caused by some task objects that are missing a done_callback. This in turn could cause exceptions to be swallowed, I **think**.

I wonder if something like this could fix it.

### DIFF
--- a/aio_pika/tools.py
+++ b/aio_pika/tools.py
@@ -42,6 +42,7 @@ def _task_done(future: asyncio.Future) -> None:
     if exc is not None:
         raise exc
 
+
 def create_task(
     func: Callable[..., Union[Coroutine[Any, Any, T], Awaitable[T]]],
     *args: Any,
--- a/aio_pika/tools.py
+++ b/aio_pika/tools.py
@@ -37,6 +37,11 @@ def iscoroutinepartial(fn: Callable[..., Any]) -> bool:
     return asyncio.iscoroutinefunction(parent)
 
 
+def _task_done(future: asyncio.Future) -> None:
+    exc = future.exception()
+    if exc is not None:
+        raise exc
+
 def create_task(
     func: Callable[..., Union[Coroutine[Any, Any, T], Awaitable[T]]],
     *args: Any,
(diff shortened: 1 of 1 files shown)


mosquito (approved):

aio_pika/tools.py
```diff
@@ -45,8 +45,15 @@ def create_task(
 ) -> Awaitable[T]:
     loop = loop or asyncio.get_event_loop()
 
+    def _task_done(future: asyncio.Future) -> None:
```
decaz, line 48: It's better to move this helper outside, - there is no need to define it for every task.

aio_pika/tools.py
```diff
@@ -45,8 +45,15 @@ def create_task(
...
+        exc = future.exception()
+        if exc:
```
decaz, line 50: I think it should be checked as following: `if exc is not None:`.

decaz (changes requested):

torarvid:
Thanks for the helpful comments, @decaz. I tried implementing both suggestions in https://github.com/mosquito/aio-pika/pull/492/commits/30c8c4f905b70b07601256b4443d4ec44fc30c24.

aio_pika/tools.py
```diff
@@ -37,6 +37,11 @@ def iscoroutinepartial(fn: Callable[..., Any]) -> bool:
     return asyncio.iscoroutinefunction(parent)
 
 
+def _task_done(future: asyncio.Future) -> None:
+    exc = future.exception()
```
decaz, line 41: This will raise `asyncio.CancelledError` in case of the `future` was cancelled. @mosquito if it's ok then we can merge this.

decaz (approved):

mosquito:
@torarvid please fix the linter warnings

aio_pika/tools.py
mosquito: It's should be raised in event loop stack and might be handled by `loop.default_exception_handler` in base case

torarvid:
@mosquito I **think** that fixed the linter error. Is it customary that I squash the commits before you merge, or will you just do that? What do you prefer? 😊

mosquito (approved):

decaz (approved):

dizballanze (approved):

mosquito:
Released as `8.2.2`

PR #493, opened 2022-09-22
Remove loop from examples

Olegt0rr:
Application developers should typically use the high-level asyncio functions, such as `asyncio.run()`, and should rarely need to reference the `loop` object or call its methods

### DIFF
--- a/docs/source/examples/master.py
+++ b/docs/source/examples/master.py
@@ -27,5 +27,4 @@ async def main() -> None:
 
 
 if __name__ == "__main__":
-    loop = asyncio.get_event_loop()
-    loop.run_until_complete(main())
+    asyncio.run(main())
--- a/docs/source/examples/pooling.py
+++ b/docs/source/examples/pooling.py
@@ -6,18 +6,16 @@ from aio_pika.pool import Pool
 
 
 async def main() -> None:
-    loop = asyncio.get_event_loop()
-
     async def get_connection() -> AbstractRobustConnection:
         return await aio_pika.connect_robust("amqp://guest:guest@localhost/")
 
-    connection_pool: Pool = Pool(get_connection, max_size=2, loop=loop)
+    connection_pool: Pool = Pool(get_connection, max_size=2)
 
     async def get_channel() -> aio_pika.Channel:
         async with connection_pool.acquire() as connection:
             return await connection.channel()
 
-    channel_pool: Pool = Pool(get_channel, max_size=10, loop=loop)
+    channel_pool: Pool = Pool(get_channel, max_size=10)
     queue_name = "pool_queue"
 
     async def consume() -> None:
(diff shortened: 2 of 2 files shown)


Olegt0rr:
Agree :)
The second commit reverted

mosquito (dismissed):

decaz (changes requested):
If we consider docs then there is also `loop` parameter 
here https://github.com/mosquito/aio-pika/blame/master/docs/source/rabbitmq-tutorial/2-work-queues.rst#L162 
and here https://github.com/mosquito/aio-pika/blob/master/docs/source/rabbitmq-tutorial/examples/6-rpc/rpc_client.py.

Olegt0rr:
> If we consider docs then there is also `loop` parameter
> here https://github.com/mosquito/aio-pika/blame/master/docs/source/rabbitmq-tutorial/2-work-queues.rst#L162
> and here https://github.com/mosquito/aio-pika/blob/master/docs/source/rabbitmq-tutorial/examples/6-rpc/rpc_client.py.

Now fixed

decaz (changes requested):
@Olegt0rr it's better to use `asyncio.get_running_loop` instead of `asyncio.get_event_loop` because the latter is deprecated now (https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop).

decaz:
@mosquito did we dropped support of 3.6? If no then @Olegt0rr should left the usage of `asyncio.get_event_loop`.

decaz:
@mosquito I asked because we still have "Python 3.6+ compatible." in README.

mosquito:
@decaz fixed

mosquito (dismissed):

docs/source/rabbitmq-tutorial/examples/6-rpc/rpc_client.py
```diff
@@ -12,16 +12,13 @@ class FibonacciRpcClient:
     connection: AbstractConnection
     channel: AbstractChannel
     callback_queue: AbstractQueue
-    loop: asyncio.AbstractEventLoop
 
     def __init__(self) -> None:
         self.futures: MutableMapping[str, asyncio.Future] = {}
-        self.loop = asyncio.get_running_loop()
+        self._loop: asyncio.AbstractEventLoop = asyncio.get_event_loop()
```
decaz, line 18: `asyncio.get_event_loop` is deprecated now (https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop).

decaz (changes requested):

docs/source/rabbitmq-tutorial/examples/6-rpc/rpc_client.py
Olegt0rr: removed loop from init, it's redundant here
we can get it directly in coro

decaz (approved):

PR #495, opened 2022-09-26
Fix memory leak

### LINKED ISSUE #494: Memory leak on channel
When declaring a queue its close callbacks are being added to close callbacks of a channel:
https://github.com/mosquito/aio-pika/blob/b0fcc56c39b329861b26bbdab271ab08f4ff8692/aio_pika/channel.py#L327

But there is no removing of these callbacks.

decaz:
Fixes #494.

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -324,7 +324,7 @@ class Channel(ChannelContext):
         )
 
         await queue.declare(timeout=timeout)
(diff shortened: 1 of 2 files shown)


mosquito (approved):

dizballanze (approved):

mosquito:
Released in 8.2.3

PR #496, opened 2022-10-03
Fix memory leaks

przemyslawfierek-silvair:
I've found two places that keep references to the Channel object methods and that prevents `Garbage Collecter` from deleting the channel from the memory. As result, after restoring the connection the `RobustConnection` restores all already closed channels.

The issue reproduction steps:
1. Start locally RabbitMQ broker instance
2. Create RobustConnection to the broker and open some RobustChannel
3. Check in a panel (i.e. RabbitMQ management) that the channel is opened
4. Close the channel
5. Restart broker
6. Check that you already opened 2 channels

### DIFF
--- a/aio_pika/channel.py
+++ b/aio_pika/channel.py
@@ -184,6 +184,9 @@ class Channel(ChannelContext):
     async def _on_close(self, closing: asyncio.Future) -> None:
         await self.close_callbacks(closing.exception())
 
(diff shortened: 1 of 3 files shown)


przemyslawfierek-silvair:
I'm working on fix

mosquito (approved):

mosquito:
Released in 8.2.4

PR #498, opened 2022-10-05
Reject all messages on close

### LINKED ISSUE #497: QueryIterator#close rejects at most 1 message
QueryIterator#close ends with these lines:
https://github.com/mosquito/aio-pika/blob/84f743a90854da1c7a8f2e95ff2be8590b4d283b/aio_pika/queue.py#L401-L412

My own conclusion is that the QueueEmpty exception should be suppressed outside the local function, let me explain why.

Scenario 1: channel is closed. Early return. 0 messages rejected. OK

Scenario 2: empty queue. QueueEmpty is thrown but suppressed, then there is an unconditional return. 0 messages rejected. OK

Scenario 3: 1 message in the queue. One message is yielded, after which we reach the unconditional return. 1 message rejected. OK

Scenario 4: 2 or more messages in the queue. One message is yielded, after which we reach the unconditional return. 1 message is rejected. NOT WHAT WE WANT

---

Removing the unconditional return would turn it into an endless loop, which is also not what we want. `QueueEmpty` should be thought of as `StopIteration` in this case. It is the mechanism to exit the loop and should therefore not be suppressed within a loop iteration.

---

I believe this should work correctly since the QueueEmpty exception exits the loop for us, which is then suppressed. 
```python
        # Reject all messages
        with contextlib.suppress(asyncio.QueueEmpty):
            while not self._amqp_queue.channel.is_closed:
                msg: IncomingMessage = self._queue.get_nowait()
                await msg.reject(requeue=True)
```

Darsstar:
Fixes #497


tests/test_pool.py
```diff
@@ -9,12 +9,13 @@
...
 async def test_simple(max_size, loop):
     counter = 0
+    lock = asyncio.Lock()
 
     async def create_instance():
         nonlocal counter
-        await asyncio.sleep(0)
-        counter += 1
-        return counter
+        async with lock:
```
mosquito, line 16: the main idea here is this block might be completely lock-free in the user code.

tests/test_amqps.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 import aio_pika
-from tests.test_amqp import TestCaseAmqp
+import tests
```
mosquito, line 7: I don't mind, just tell me why?

aio_pika/queue.py
```diff
@@ -398,18 +398,11 @@ async def close(self, *_: Any) -> Any:
...
-                return None
-
         # Reject all messages
-        msg: IncomingMessage
-        for msg in queue_tail(self._amqp_queue.channel):
-            await msg.reject(requeue=True)
+        with contextlib.suppress(asyncio.QueueEmpty):
+            while not self._amqp_queue.channel.is_closed:
+                msg: IncomingMessage = self._queue.get_nowait()
+                await msg.reject(requeue=True)
```
mosquito, line 405: I will be really happy if you try my method with `.nack(requeue=True, multiple=True)` only for the last message.

README.rst
```diff
@@ -254,7 +254,7 @@ _____________
 
 .. code-block:: bash
 
-    docker run -d -p 5672:5672 -p 15672:15672 rabbitmq:3-management
+    docker run -d -p 5671:5671 -p 5672:5672 -p 15671:15671 -p 15672:15672 mosquito/aiormq-rabbitmq
```
mosquito, line 257: Nice fix. Don't you want to break this into escaped multiline string?

mosquito (changes requested):

tests/test_amqps.py
Darsstar: As the commit message mentions: "Don't import TestCaseAmqp into global scope since pytest would detect it and run most tests in this file twice."
Less tests finishes faster.

tests/test_pool.py
Darsstar: I removed the lock and made the time while holding and instance conditional on whether `max_size` has been reached or not. The code will sleep longer it has not been and sleeps shorter when it has been.

(My PC only got up to 3 instances when it was sleeping 0.01 seconds.)

README.rst
Darsstar: I do not want to. But I will if you insist.

aio_pika/queue.py
Darsstar: Done

README.rst
mosquito: I don't insist. I just wanted to see if anyone else wants to see this in a few lines.

mosquito (approved):

README.rst
Alviner, line 257: what's wrong with official docker image?

Darsstar: Nothing, it just isn't the image the make file uses. This just updates the contributor documentation to match what happens in Github Actions.

tests/test_pool.py
```diff
@@ -23,13 +23,14 @@ async def getter():
 
         async with pool.acquire() as instance:
             assert instance > 0
-            await asyncio.sleep(0.01)
-            return counter
+            await asyncio.sleep(1 if counter < max_size else 0)
```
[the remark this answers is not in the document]
Darsstar: Because I don't see the point in these tests taking 4 minutes and 27 second (266.93s) (unconditional 1 second) when they can take 3.21s instead (conditionally 1 or 0 seconds).

The longer the sleep time the more likely it is the test won't produce a false negative. 0.01 unconditionally produces false negatives on my PC because it is fast enough that it only needs to create 3 instances during the entire test.

The commit which changes that also has the same explanation in the commit message but worded differently.

README.rst
Darsstar: Well, it is wrong in the sense that it is not entirely what the test suite expects.

decaz (approved):

Alviner (approved):

mosquito:
Released in 8.2.4

PR #507, opened 2022-11-07
Fix timeout logging on QueueIterator

pixelneo:
A wrong information (the default timeout) is being printed (instead of the supplied one in kwargs).


mosquito (dismissed):

pixelneo:
Hello @mosquito, could you please also approve the workflow (it's required before merge)?

decaz:
@pixelneo please resolve the issue with the lint check.

mosquito (approved):

PR #511, opened 2022-12-27
Rework reconnect logic

mosquito:


tests/test_amqp.py
```diff
@@ -1597,7 +1601,7 @@ async def test_connection_close(
         try:
             with pytest.raises(aio_pika.exceptions.ChannelPreconditionFailed):
                 msg = Message(bytes(shortuuid.uuid(), "utf-8"))
-                msg.delivery_mode = 8
+                msg.delivery_mode = 8       # type: ignore
```
Alviner, line 1604: what's happened here? why mypy is struggling?

Alviner (approved):

Alviner (dismissed):

tests/test_amqp.py
mosquito: Because it's IntEnum

Alviner (approved):

Alviner (dismissed):

aio_pika/abc.py
```diff
@@ -114,20 +115,36 @@ async def __aexit__(
...
-    HeadersValue,
-    Set[HeadersValue],
-    Tuple[HeadersValue, ...],
-    FrozenSet[HeadersValue],
-]
-HeadersType = MutableMapping[str, HeadersPythonValues]
+HeadersType = Dict[str, FieldValue]
+
+
+class MessageInfo(TypedDict, total=False):
```
decaz, line 121: Why not dataclass?

aio_pika/abc.py
```diff
@@ -257,7 +265,17 @@ def __init__(
         arguments: Arguments,
         passive: bool = False,
     ):
-        raise NotImplementedError
+        raise NotImplementedError(
+            dict(
```
decaz, line 269: Why do we need this data here? It's the only place where `NotImplementedError` is being raised with data..

aio_pika/abc.py
mosquito: dataclass with total=False is a very strange solution. Message info it's a way for print message to logs etc. The dictionary is a completely relevant for this.

aio_pika/abc.py
mosquito: This code is a hack for linters it's never being called in the runtime

pyproject.toml
```diff
@@ -0,0 +1,107 @@
+[tool.poetry]
+name = "aio-pika"
+version = "9.0.0"
+description = "Wrapper for the aiormq for asyncio and humans"
```
decaz, line 4: Wrapper **over** aiormq ... would be better, I guess.

aio_pika/robust_connection.py
```diff
@@ -95,11 +97,64 @@ async def __connection_attempt(
...
+                await asyncio.sleep(self.reconnect_interval)
+            except Exception:
+                log.exception(
+                    "Reconnect attempt failed %s. "
+                    "Retrying after %r seconds.",
+                    self, self.reconnect_interval,
+                )
+                await asyncio.sleep(self.reconnect_interval)
+
+            await asyncio.sleep(self.reconnect_interval)
```
decaz, line 153: I guess two sleeps above are redundant, are they?

mosquito: Nope previous one in the except block

pyproject.toml
mosquito: around?

decaz: Yes, it's suitable 👍🏻

aio_pika/robust_connection.py
decaz: Yes, but this sleep will run after these `except` blocks.

mosquito: You right

decaz (approved):
@mosquito great work, thank you!

PR #521, opened 2023-02-14
Do not nack if consumer is no_ack in QueueIterator

jkr78:
An exception is raised if `QueueIterator` is created with the consumer flag `no_ack`.
Added a check to ensure `nack` is not getting called when `QeueuIterator` is closed and `consumer` has `no_ack` option set to `True`.

### DIFF
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -416,7 +416,7 @@ class QueueIterator(AbstractQueueIterator):
 
             if self._consume_kwargs.get("no_ack", False):
                 log.warning(
-                    "Message %r lost for consumer with no_ack",
+                    "Message %r lost for consumer with no_ack %r",
                     msg,
                     self,
                 )
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -403,13 +403,27 @@ class QueueIterator(AbstractQueueIterator):
             while True:
                 msg = self._queue.get_nowait()
         except asyncio.QueueEmpty:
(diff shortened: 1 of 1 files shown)


aio_pika/queue.py
```diff
@@ -403,7 +403,12 @@ async def close(self, *_: Any) -> Any:
             while True:
                 msg = self._queue.get_nowait()
         except asyncio.QueueEmpty:
-            if msg is not None and not self._amqp_queue.channel.is_closed:
+            nack_ok = (
```
mosquito, line 406: I would like to see a log.warning that these messages have been received, not processed and will be lost.

mosquito (changes requested):

aio_pika/queue.py
```diff
@@ -403,13 +403,27 @@ async def close(self, *_: Any) -> Any:
...
+                log.warning(
+                    "Message %r lost when queue iterator %r channel closed",
+                    msg,
+                    self,
+                )
+                return
+
+            if self._consume_kwargs.get("no_ack", False):
+                log.warning(
+                    "Message %r lost for consumer with no_ack",
```
mosquito, line 419: forgot `%r`

mosquito (dismissed):

mosquito (commented):
Satisfy pylama

Alviner (approved):

mosquito:
@jkr78 your fixed released in 9.0.2. Thank you for your contribution.

PR #526, opened 2023-02-28
Avoid implicitly depending on setuptools

zware:
Currently there's an implicit dependency on `pkg_resources` from `setuptools` to get the version of the package.  This makes the dependency explicit for Python 3.7 and avoids it entirely for 3.8+ which have `importlib.metadata` available.


Alviner (approved):

zware:
Thanks!

PR #533, opened 2023-03-27
Feature/wait reopen channel state

### LINKED ISSUE #505: Reconnect does not work with version 8.2.4 but working with version 8.2.0
Hello!

First of all, thank you for your great work! I probably found a bug with aio-pika 8.2.4 version. I'm using windows, docker image with rabbitmq, python 3.9.

Code example for reproducing:

```python
import asyncio

import aio_pika
from aio_pika import ExchangeType
from aio_pika.abc import AbstractIncomingMessage
import aiorun


async def _consume(message: AbstractIncomingMessage) -> None:
    async with message.process(requeue=True, ignore_processed=True):
        print(
            f"[x] {message.routing_key}"
            f"[x] Decoded {message.body.decode()}"
        )


async def _check_tasks() -> None:
    while True:
        tasks = asyncio.all_tasks()
        print(f"Number of tasks: {len(tasks)}")
        for task in tasks:
            print(f"Task: {task.get_name()} {task.get_coro()}")
        print("================================")
        await asyncio.sleep(10)


async def _receive() -> None:
    connection = await aio_pika.connect_robust(
        "amqp://guest:guest@127.0.0.1/"
    )
    channel = await connection.channel()
    await channel.set_qos(prefetch_count=20)
    exchange = await channel.declare_exchange(
        "broker_publish",
        ExchangeType.TOPIC,
    )
    queue = await channel.declare_queue(
        "broker_events",
        durable=False,
        exclusive=False,
        auto_delete=False,
    )
    await queue.bind(exchange, routing_key="#")
    await queue.consume(_consume)


async def main() -> None:
    asyncio.create_task(_receive())
    asyncio.create_task(_check_tasks())


if __name__ == "__main__":
    aiorun.run(
        main(),
        stop_on_unhandled_errors=False,
    )

```

Steps:

1)  Run the script (with aio-pika 8.2.4). You will see 7 running tasks (I think on Linux it would be 6).
2) Publish message to queue. You will see this message in terminal.
3) Restart docker container with RabbitMQ.
4) Thera are only 6 running tasks after reconnect (Channel._reader is missing)
5) Try to publish another message. No result in terminal. In RMQ GUI no consumers related to this queue.

6) Repeat steps 1-5 with aio-pika. Number of tasks will remain after reconnect and the script will receive message.

I can provide more information if needed. Thank you very much!

mosquito:
Fixes #505


Alviner (approved):

aio_pika/robust_channel.py
```diff
@@ -57,51 +60,73 @@ def __init__(
...
         async with self.__restore_lock:
-            self._connection = connection
+            if self.__restored:
+                return
+
             await self.reopen()
+            self.__restored = True
+
+    async def __close_callback(self, _: Any, exc: BaseException) -> None:
+        if isinstance(exc, asyncio.CancelledError):
```
tzoiker, line 91: 🤔

aio_pika/channel.py
```diff
@@ -119,45 +122,55 @@ async def close(
...
 
-        if self.is_closed:
-            raise aiormq.exceptions.ChannelInvalidStateError(
-                "Channel has been closed",
-            )
+        return self._channel.channel
 
+    @property
+    def channel(self) -> aiormq.abc.AbstractChannel:
+        warnings.warn("This property is deprecated, do not use this anymore.")
```
Pavkazzz: lets warn `DeprecationWarning`

tzoiker (dismissed):

aio_pika/robust_exchange.py
```diff
@@ -45,11 +47,15 @@ def __init__(
         self._bindings = {}
         self.__restore_lock = asyncio.Lock()
 
-    async def restore(self, channel: aiormq.abc.AbstractChannel) -> None:
+    async def restore(self, channel: Any = None) -> None:
+        if channel is not None:
+            warnings.warn(
+                "Channel argument will be ignored because you "
+                "no need to pass this anymore.",
```
decaz, line 54: ... **don't** need ...

aio_pika/robust_queue.py
```diff
@@ -54,8 +55,14 @@ def __init__(
         self._consumers = {}
         self._bindings = {}
 
-    async def restore(self, channel: aiormq.abc.AbstractChannel) -> None:
-        self.channel = channel
+    async def restore(self, channel: Any = None) -> None:
+        if channel is not None:
+            warnings.warn(
+                "Channel argument will be ignored because you "
+                "no need to pass this anymore.",
```
decaz, line 62: ... **don't** need ...

decaz (changes requested):

aio_pika/robust_channel.py
```diff
@@ -141,6 +178,7 @@ async def declare_exchange(
         robust: bool = True,
     ) -> AbstractRobustExchange:
         await self._connection.ready()
```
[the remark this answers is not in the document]
mosquito: `self._connection` might be in not initialised state when `self.ready` calling

Alviner (changes requested):

aio_pika/patterns/master.py
```diff
@@ -143,17 +141,18 @@ def deserialize(self, data: bytes) -> Any:
 
     @classmethod
     async def execute(
-        cls, func: Callable[..., Awaitable[T]], kwargs: Any,
+        cls, func: CallbackType, kwargs: Any,
     ) -> T:
         kwargs = kwargs or {}
 
         if not isinstance(kwargs, dict):
+            logging.error("Bad kwargs %r received for the %r", kwargs, func)
```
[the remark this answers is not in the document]
Alviner: I mean that we don't use 'log' but 'logging' here

Alviner (dismissed):

phillipuniverse:
@mosquito question about this with the removed `connection` construct.

In the aio_pika OpenTelemetry instrumentation [located here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a6797541721af72e6feab400e24ceaabc08cb31e/instrumentation/opentelemetry-instrumentation-aio-pika/src/opentelemetry/instrumentation/aio_pika/span_builder.py#L51-L64) there are a few fallbacks to getting the active URL from the channel. The connection url is being retrieved from the channel since that's all that is really available at publish time, which otel [decorates here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a6797541721af72e6feab400e24ceaabc08cb31e/instrumentation/opentelemetry-instrumentation-aio-pika/src/opentelemetry/instrumentation/aio_pika/publish_decorator.py#L41-L44).

I'm wondering if you have some advice on what the correct thing to do in this case would be to support the changes in 9.1 while still supporting 7.x, 8.x and 9.1+?

I would hate to drop down to use `channel._connection` (although that will work _most_ of the time), do you think change the method to async and use the `get_underlay_channel()`? Like this:

```python
     async def set_channel(self, channel: AbstractChannel):
        # aio_rmq 9.1 removed the connection attribute from the abstract listings
        if hasattr(channel, "get_underlay_channel"):
            connection = (await channel.get_underlay_channel()).connection
        else:
            connection = channel.connection
        if getattr(connection, "connection", None):
            # aio_rmq 7
            url = connection.connection.url
        else:
            # aio_rmq 8
            url = connection.url
        self._attributes.update(
            {
                SpanAttributes.NET_PEER_NAME: url.host,
                SpanAttributes.NET_PEER_PORT: url.port,
            }
        )
```

Or would it be better in your opinion to keep this sync and check for `channel._connection` like this:

```python
     async def set_channel(self, channel: AbstractChannel):
        # aio_rmq 9.1 removed the connection attribute from the abstract listings
        if hasattr(channel, "_connection"):
            connection = channel._connection
        else:
            connection = channel.connection
        if getattr(connection, "connection", None):
            # aio_rmq 7
            url = connection.connection.url
        else:
            # aio_rmq 8
            url = connection.url
        self._attributes.update(
            {
                SpanAttributes.NET_PEER_NAME: url.host,
                SpanAttributes.NET_PEER_PORT: url.port,
            }
        )
```

PR #541, opened 2023-05-08
Queue.declare() passive parameter not recognized

Aurvandill:
In the `Queue` class the Function `declare` takes an argument called `passive` according to docstrings.
Currently it does not.
This PR adds it as an optional parameter.
if no argument is given it will default to `self.passive`

### DIFF
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -66,7 +66,7 @@ class Queue(AbstractQueue):
         )
 
     async def declare(
-        self, timeout: TimeoutType = None,
+        self, timeout: TimeoutType = None, passive: Optional[bool] = None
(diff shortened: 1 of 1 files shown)


Aurvandill:
@mosquito any chance that you will take a look at this?
with best regards
Aurvandill

decaz:
@Aurvandill `passive` parameter of a queue should be passed in `channel.declare_queue` method. After that `Queue.declare` inherits its value therefore there is inaccuracy in docstring of this method. Feel free to remove this parameter from docstring within current PR.

mosquito (approved):

mosquito:
So miss clicked actually 😅

Aurvandill:
thanks for replying
@decaz thanks for the explanation

PR #542, opened 2023-05-10
Improve type hints for queue.get()

asvetlov:
The master assumes that the message should be checked for not-none even if `fail=True` (default) and None is never returned.
```
msg = await queue.get()
assert msg is not None
```

The proposal avoids the need for such asserts by using `typing.Literal` and method overloads.

### DIFF
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -1,5 +1,6 @@
 import asyncio
 import dataclasses
+import sys
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
 from enum import Enum, IntEnum, unique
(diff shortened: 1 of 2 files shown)


Alviner (dismissed):

mosquito (approved):

decaz (approved):
@asvetlov thank you!

Alviner (approved):

asvetlov:
@mosquito sorry, I didn't find a way to execute linters locally.
Next time I know what to do :)

mosquito:
@asvetlov just run `poetry run pylama`.

asvetlov:
Ok, thanks for letting me know

PR #549, opened 2023-05-28
Documentation fixes

### LINKED ISSUE #548: Documentation issues
I think I encountered some issues in the RabbitMQ tutorial. (I haven't tried the library yet, though)

* https://aio-pika.readthedocs.io/en/latest/rabbitmq-tutorial/1-introduction.html#hello-world

> RabbitMQ libraries
>
> RabbitMQ speaks AMQP 0.9.1, which is an open, general-purpose protocol for messaging. There are a number of clients for RabbitMQ in [many different languages](https://www.rabbitmq.com/devtools.html). In this tutorial series we’re going to use [aio-pika](https://github.com/mosquito/aio-pika), **which is the Python client recommended by the RabbitMQ team**. To install it you can use the [pip](https://pip.pypa.io/en/stable/quickstart/) package management tool.

Is that really the case? I couldn't find anything "official". They only ever mention `pika` on their website.

* https://aio-pika.readthedocs.io/en/latest/rabbitmq-tutorial/2-work-queues.html#preparation
> We don’t have a real-world task, like images to be resized or pdf files to be rendered, so let’s fake it by just pretending we’re busy - by using the time.sleep() function.

The `time.sleep()` function is not used in that section. Only much later in "Message acknowledgment" is `asyncio.sleep` used.

> Our old receive.py script also requires some changes: **it needs to fake a second of work for every dot in the message body**. It will pop messages from the queue and perform the task, so let’s call it tasks_worker.py:
> ```
> async def on_message(message: AbstractIncomingMessage) -> None:
>     async with message.process():
>         print(f" [x] Received message {message!r}")
>         print(f"     Message body is: {message.body!r}")
> ```

This does not do any sleeping.

* https://aio-pika.readthedocs.io/en/latest/rabbitmq-tutorial/3-publish-subscribe.html#temporary-queues

> Secondly, once we disconnect the consumer the queue should be deleted. There’s an exclusive flag for that:
> ```
> queue = await channel.declare_queue(exclusive=True)
> ```

I'm not sure if aio-pika uses a different wording than pika or rabbitmq, but `exclusive` is not the word for "automatic deletion". I thought `exclusive` means, that only a single consumer can connect to a queue. There is also an [`auto_delete`](https://aio-pika.readthedocs.io/en/latest/apidoc.html#aio_pika.Channel.declare_queue) parameter, so I assume that is an error.

* Patterns and helpers

I feel like there needs to be more detailed documentation of what the patterns actually do. They are written in a way that it seems like they can only be used between two `aio-pika` implementations and there is no possible interoperability between other AMQP implementations possible?

From reading the documentation it feels like `aio-pika` is a lot more ergonomic for asynchronous usage than `pika`, but the fact that it is not based on the "official recommended implementation" makes it a lot harder to argue for. Can you explain, why it was not possible to keep this as a wrapper around `pika` and you had to go for your own implementation?

MaPePeR:
Fixes #548 

I tried fixing some minor issues and continuity errors in the documentation.

### DIFF
--- a/docs/source/rabbitmq-tutorial/examples/2-work-queues/new_task_initial.py
+++ b/docs/source/rabbitmq-tutorial/examples/2-work-queues/new_task_initial.py
@@ -1,0 +1,29 @@
+import asyncio
+import sys
+
+from aio_pika import Message, connect
+
+
+async def main() -> None:
+    # Perform connection
+    connection = await connect("amqp://guest:guest@localhost/")
+
+    async with connection:
+        # Creating a channel
+        channel = await connection.channel()
+
+        message_body = b" ".join(
+            arg.encode() for arg in sys.argv[1:]
+        ) or b"Hello World!"
+
+        # Sending the message
+        await channel.default_exchange.publish(
+            Message(message_body),
+            routing_key="hello",
+        )
+
+        print(f" [x] Sent {message_body!r}")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/docs/source/rabbitmq-tutorial/1-introduction.rst
+++ b/docs/source/rabbitmq-tutorial/1-introduction.rst
@@ -104,8 +104,7 @@ Producer sends messages to the "hello" queue. The consumer receives messages fro
 
     RabbitMQ speaks AMQP 0.9.1, which is an open, general-purpose protocol for messaging.
     There are a number of clients for RabbitMQ in `many different languages`_.
-    In this tutorial series we're going to use `aio-pika`_,
-    which is the Python client recommended by the RabbitMQ team.
+    In this tutorial series we're going to use `aio-pika`_.
     To install it you can use the `pip`_ package management tool.
 
     .. _many different languages: https://www.rabbitmq.com/devtools.html
--- a/docs/source/rabbitmq-tutorial/2-work-queues.rst
+++ b/docs/source/rabbitmq-tutorial/2-work-queues.rst
@@ -65,7 +65,7 @@ We will slightly modify the send.py code from our previous example, to allow arb
 messages to be sent from the command line. This program will schedule tasks to our work
 queue, so let's name it *new_task.py*:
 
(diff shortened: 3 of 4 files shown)


docs/source/rabbitmq-tutorial/examples/2-work-queues/new_task_initial.py
```diff
@@ -0,0 +1,29 @@
...
+            arg.encode() for arg in sys.argv[1:]
+        ) or b"Hello World!"
+
+        # Sending the message
+        await channel.default_exchange.publish(
+             Message(message_body),
+             routing_key="hello",
+        )
+
+        print(f" [x] Sent {message!r}")
```
mosquito, line 25: `message_body`?

docs/source/rabbitmq-tutorial/examples/2-work-queues/new_task_initial.py
```diff
@@ -0,0 +1,29 @@
...
+    async with connection:
+        # Creating a channel
+        channel = await connection.channel()
+
+        message_body = b" ".join(
+            arg.encode() for arg in sys.argv[1:]
+        ) or b"Hello World!"
+
+        # Sending the message
+        await channel.default_exchange.publish(
```
mosquito, line 20:
```python
confirmation = await channel.default_exchange.publish(
```

mosquito (changes requested):

docs/source/rabbitmq-tutorial/examples/2-work-queues/new_task_initial.py
MaPePeR: Then `pylama` will scream because of the unused variable. What shall I do with the `confirmation`?
I tried to do something useful with it like reading the `delivery_tag`, but this is how i encountered the unexpected `DeliveredMessage` by accident. (mosquito/aiormq#183)

mosquito (approved):

mosquito:
@MaPePeR excellent, merged.

PR #560, opened 2023-07-04
use fork friendly random robust queue generation way

mosquito:


decaz (changes requested):
I guess `RobustQueue._rnd_gen` attribute can be removed then.

decaz (approved):

PR #561, opened 2023-07-05
Fix `See also` Sphinx headers

Lancetnik:
My bad, I am not a .rst user, thus made a mistake

For some reason, github and sphinx are using the different .rst parsers (or differ from [this guide](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings)) and the same README markup represents wrong in `aio-pika` sphinx doc: there no any `See also` sections subtitles.

So I fix It and now we have the following view:
image

Instead the current:
image

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -190,10 +190,10 @@ Thanks for contributing
 
 
 See also
-==========
+++++++++
 
-`aiormq <https://github.com/mosquito/aiormq>`_
-----------------------------------------------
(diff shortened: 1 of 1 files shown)


mosquito (approved):

Lancetnik:
Wow, It was fast

mosquito:
@Lancetnik actually was thinking about similar fix but planned do it evening

PR #569, opened 2023-07-26
URL params passing to aiormq

### LINKED ISSUE #567: Issues with HA RabbitMQ cluster
Hi,
I tried making a detailed issue but I don't know my way around aio-pika and aiormq enough, so I'll make it simple.

**setup**
I run rabbitmq in a cluster and I use quorum queues. One rabbit instance might become unavailable at any time.
I expect `RobustConnection` to handle reconnects to any rabbit replicas automagically (I'm setup like the example `Asynchronous message processing`).

Seems like aiormq keeps a persistent connection open with a single rabbit replica. When this replica goes down I get these **issues**:
1. An exception is created each time, sending many false alerts my way 
    https://github.com/mosquito/aiormq/blob/f04f0cf7f1972e14abd4fefc90218770f335e239/aiormq/connection.py#L558
2. Occasional `Task was destroyed but it is pending!` (#432) seem to happen when reading a message during rmq shutdown (it happens before `Unexpected connection close` is printed).
    ```
    Task was destroyed but it is pending!
    task: <Task pending name='Task-31' coro=<Connection._on_reader_done.<locals>.close_writer_task() running at /usr/local/lib/python3.10/site-packages/aiormq/connection.py:506> wait_for=<Future finished result=None>>
    task: <Task pending name='Task-38' coro=<OneShotCallback.__task_inner() running at /usr/local/lib/python3.10/site-packages/aio_pika/tools.py:230>>
    ```
3. Occasional `DeliveryError` from `exchange.publish(message, MY_ROUTING_KEY)`, seems to happen in tandem with `Task was destroyed but it is pending!` (before `Unexpected connection close`).

---

Is there a magic bullet to solve all of this ?
Otherwise, how can I help solve these issues ?

mosquito:
Fixes #567


decaz (approved):

Alviner (approved):

.editorconfig
```diff
@@ -8,13 +8,11 @@ trim_trailing_whitespace = true
 
 [*.{py,yml}]
 indent_style = space
+max_line_length = 80
```
decaz, line 11: Maybe 79 to be consistent with PEP 8?

aio_pika/abc.py
```diff
@@ -692,10 +694,29 @@ async def close(self, exc: Optional[aiormq.abc.ExceptionType]) -> Any:
             await self.close_callback.wait()
 
 
+@dataclass
+class ConnectionParameter:
+    name: str
+    parser: Callable[[str], Any]
+    default: Optional[str] = None
+    kwarg: bool = False
```
decaz, line 702: I think it would be better to use "is_" prefix with boolean variable names: "is_kwarg".

aio_pika/abc.py
```diff
@@ -692,10 +694,29 @@ async def close(self, exc: Optional[aiormq.abc.ExceptionType]) -> Any:
...
+        if value is None:
+            return self.default
+        try:
+            return self.parser(value)
+        except ValueError:
+            return self.default
+
+
 class AbstractConnection(PoolInstance, ABC):
+    KWARGS_TYPES: Tuple[ConnectionParameter, ...]
```
decaz, line 714: What to you think about renaming this class variable to "PARAMETERS"?

aio_pika/abc.py
```diff
@@ -692,10 +694,29 @@ async def close(self, exc: Optional[aiormq.abc.ExceptionType]) -> Any:
...
+
     close_callbacks: CallbackCollection
     connected: asyncio.Event
     transport: Optional[UnderlayConnection]
+    kwargs: Mapping[str, Any]
```
decaz, line 719: And renaming this to "parameters"?

aio_pika/connection.py
```diff
@@ -46,8 +58,14 @@ async def close(
     @classmethod
     def _parse_kwargs(cls, kwargs: Dict[str, Any]) -> Dict[str, Any]:
```
decaz, line 59: Then this method also can be renamed too:
```python
@classmethod
def _parse_parameters(cls, parameters: Dict[str, Any]) -> Dict[str, Any]:
    ...
```

decaz (changes requested):
The main idea is to use name "parameter" instead of "kwarg".

.editorconfig
mosquito: it just a ruller for the editor, stict check in the liter.

decaz (approved):

PR #571, opened 2023-08-05
Fix reopen of robust channel after close

### LINKED ISSUE #570: Robust channel stucks after closing and reopening
If you close and reopen robust channel then any action on that channel (declaring a queue for example) will await channel readiness for infinite time.

decaz:
Fixes #570.


mosquito (approved):

Alviner (approved):

mosquito (changes requested):
So, we must wait channel will be declared before sending some channel-related frames.

mosquito (approved):
By bad. This should be ok

PR #573, opened 2023-08-12
Fix bug with RPC when handling on_close with a RobustConnection

CodeCorrupt:
When handing on_close, futures were not properly popped from the futures list, causing an asyncio.exceptions.InvalidStateError when the connection is re-established and the response received.

### DIFF
--- a/aio_pika/patterns/rpc.py
+++ b/aio_pika/patterns/rpc.py
@@ -99,7 +99,9 @@ class RPC(Base):
         self.consumer_tags: Dict[Callable[..., Any], ConsumerTag] = {}
         self.host_exceptions = host_exceptions
 
-    def __remove_future(self, correlation_id: str) -> Callable[[asyncio.Future], None]:
+    def __remove_future(
+        self, correlation_id: str
+    ) -> Callable[[asyncio.Future], None]:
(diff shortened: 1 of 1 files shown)


CodeCorrupt:
Alternatively, the futures could be popped by changing [this loop](https://github.com/mosquito/aio-pika/blob/c8e33c58ff7b81c6eeff58fcaf8c44ded82f9bce/aio_pika/patterns/rpc.py#L174-L178) to use `enumerate(self.futures)`, then pop using the `correlation_id`, and then just deleting the [`self.futures.pop(str(id(future)), None)`](https://github.com/mosquito/aio-pika/blob/c8e33c58ff7b81c6eeff58fcaf8c44ded82f9bce/aio_pika/patterns/rpc.py#L104C9-L104C48) line. But the PR I submitted seemed like a more general solution to me

mosquito (changes requested):

mosquito (approved):

decaz (approved):

mosquito:
@CodeCorrupt I was thinking, the `on_close` is called quite rarely, maybe it makes sense to copy the whole sequence there so I don't have to do it every time here?

mosquito:
@CodeCorrupt released in 9.2.2

Issue #577, opened 2023-09-07
RobustChannel is not restoring after "Connection was stuck" error.


aozupek:
aio-pika==9.2.2
aiormq==6.7.7
RabbitMQ: 3.12.2 in local docker container

Related issues: https://github.com/mosquito/aiormq/issues/165, #575, #563

This problem arises with the "connection was stuck" bug (https://github.com/mosquito/aiormq/issues/165) in [aiormq](https://github.com/mosquito/aiormq). If I manually restart my RabbitMQ from docker or just toggle off/on the network of my docker container via `docker network disconnect/connect bridge rabbitmq-server` command, everything works as expected. All of the connections, channels, queues and exchanges are being restored sucessfully. However when the "connection was stuck" error occurs, only the connection is getting restored and channels remain closed. Since RobustChannel is responsible of restoring the queues and exchanges, all of them are remain closed too.

Reproducing the "connection was stuck" bug is not easy as it takes a period of time without any operation. Sometimes it may take 15 minutes and sometimes 4-5 hours. I've tried to change the heartbeat duration, let my computer sleep for a few minutes/hours, keep it open for hours without any operations, etc... but unfortunately I couldn't find a solid method to reproduce the bug.

Here are the logs when I catched it last time (heartbeat was 1):

```
DEBUG:aiormq.connection:Received frame <pamqp.heartbeat.Heartbeat object at 0x10327d1d0> in channel #0 weight=8 on <Connection: "amqp://guest:******@localhost:5672//?heartbeat=1" at 0x103262030>
DEBUG:aiormq.connection:Received frame <pamqp.heartbeat.Heartbeat object at 0x10327fe90> in channel #0 weight=8 on <Connection: "amqp://guest:******@localhost:5672//?heartbeat=1" at 0x103262030>
DEBUG:aiormq.connection:Prepare to send ChannelFrame(payload=b'\x08\x00\x00\x00\x00\x00\x00\xce', should_close=False, drain_future=None)
DEBUG:aiormq.connection:Received frame <pamqp.heartbeat.Heartbeat object at 0x10327d1d0> in channel #0 weight=8 on <Connection: "amqp://guest:******@localhost:5672//?heartbeat=1" at 0x103262030>
DEBUG:aiormq.connection:Received frame <pamqp.heartbeat.Heartbeat object at 0x10327fe90> in channel #0 weight=8 on <Connection: "amqp://guest:******@localhost:5672//?heartbeat=1" at 0x103262030>
[...]

PR #578, opened 2023-09-10
Fix restore bug of RobustChannel

aozupek:
See #577


decaz (approved):
Thank you!

mosquito (approved):
@aozupek cool job, thank you for your research on the problem.

mosquito:
released as [aio-pika==9.2.3](https://pypi.org/project/aio-pika/9.2.3/)

Issue #589, opened 2023-10-16
RuntimeError: Set changed size during iteration (in aio-pika 9.3.0)


dvarrazzo:
Hello,

I received Sentry issues from a live server, which, if I understand correctly, happened in a moment in which a misconfigured firewall closed the door in front of an aio-pika client, and the k8s health check was trying to connect.

The error is `RuntimeError: Set changed size during iteration` on this line:

https://github.com/mosquito/aio-pika/blob/0e5db7fe33eb07daa9458631376b5ce3b2422cad/aio_pika/robust_connection.py#L98

Unfortunately I can't work more on it in order to provide anything reproducible. However maybe you want to make a copy of this set before iterating it?

```diff
--- aio_pika/robust_connection.py.bak	2023-10-16 20:55:57.867441312 +0200
+++ aio_pika/robust_connection.py    	2023-10-16 20:56:12.163364473 +0200
@@ -95,7 +95,7 @@
             raise RuntimeError("No active transport for connection %r", self)
 
         try:
-            for channel in self.__channels:
+            for channel in list(self.__channels):
                 try:
                     await channel.restore()
                 except Exception:
```

aio-pika version: 9.3.0, Python version: 3.10.12. I found #388 but it claims to having been fixed in >=8, so this might be a different instance.

<details>
<summary>Full stack trace</summary>

```
CancelledError: null
  File "apps_mq/mq.py", line 265, in on_message
    await message.ack()
  File "aio_pika/message.py", line 476, in ack
    await self.channel.basic_ack(
  File "aiormq/channel.py", line 556, in basic_ack
    await drain_future

CancelledError: null
  File "aio_pika/queue.py", line 37, in consumer
    return await create_task(callback, message)

CancelledError: null
  File "aiormq/abc.py", line 44, in __inner
    return await self.task

RuntimeError: Set changed size during iteration
  File "apps_mq/mq.py", line 279, in check_health
    await asyncio.gather(test_sender(), test_receiver())
  File "asyncio/tasks.py", line 650, in _wrap_awaitable
    return (yield from awaitable.__await__())
  File "aiormq/abc.py", line 46, in __inner
    raise self._exception from e
  File "asyncio/tasks.py", line 650, in _wrap_awaitable
    return (yield from awaitable.__await__())
  File "aiormq/abc.py", line 46, in __inner
    raise self._exception from e
  File "apps_mq/mq.py", line 268, in test_receiver
    await queue.consume(on_message)
[...]

PR #590, opened 2023-10-16
Fix possible error iterating on connection channels during concurrent update

dvarrazzo:
Fix for the crash reported in #589 unfortunately not tested to verify if the fix is complete.

### DIFF
--- a/aio_pika/robust_connection.py
+++ b/aio_pika/robust_connection.py
@@ -95,7 +95,9 @@ class RobustConnection(Connection, AbstractRobustConnection):
             raise RuntimeError("No active transport for connection %r", self)
 
         try:
-            for channel in self.__channels:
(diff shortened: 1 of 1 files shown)


aio_pika/robust_connection.py
```diff
@@ -95,7 +95,9 @@ async def _on_connected(self) -> None:
             raise RuntimeError("No active transport for connection %r", self)
 
         try:
-            for channel in self.__channels:
+            # Make a copy of the channels to iterate on to guard from
+            # concurrent updates to the set.
+            for channel in list(self.__channels):
```
decaz, line 100: It would be better to use a tuple instead of a list due to execution speed and lesser memory consumption.

mosquito (approved):

PR #598, opened 2023-11-13
Define empty __slots__ in base classes

anton-ryzhov:
I've noticed that `__slots__` are defined in `Message`-classes, but they were not working as expected.

> When inheriting from a class without __slots__, the __dict__ and __weakref__ attribute of the instances will always be accessible.

https://docs.python.org/3/reference/datamodel.html?#object.__slots__

So all parent classes must also have `__slots__` to disable `__dict__` creation. That's what I did in this PR

### DIFF
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -140,6 +140,8 @@ class MessageInfo(TypedDict, total=False):
 
 
 class AbstractMessage(ABC):
+    __slots__ = ()
+
     body: bytes
     body_size: int
     headers: HeadersType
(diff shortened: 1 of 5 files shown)


mosquito (approved):

decaz (changes requested):
@anton-ryzhov can you add `__slots__` for parent classes of `RobustQueue` which also has this attribute, please?

anton-ryzhov:
Done

mosquito (approved):

decaz (approved):

PR #610, opened 2023-12-28
Do not use deprecated datetime.utcnow and datetime.utcfromtimestamp

### LINKED ISSUE #609: Deprecation warnings about datetime.utcnow and datetime.utcfromtimestamp
Method [datetime.utcnow](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcnow) is deprecated since version 3.12:
```
DeprecationWarning: datetime.datetime.utcnow() 
is deprecated and scheduled for removal in a future version. 
Instead, Use timezone-aware objects to represent datetimes in UTC: 
datetime.datetime.now(datetime.UTC).
```
Method [datetime.utcfromtimestamp](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcfromtimestamp) is deprecated since version 3.12:
```
DeprecationWarning: datetime.datetime.utcfromtimestamp() 
is deprecated and scheduled for removal in a future version. 
Use timezone-aware objects to represent datetimes in UTC: 
datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```

decaz:
This change can be breaking because timestamp of incoming message will now have timezone info (UTC by default).

Fixes #609.

### DIFF
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -38,7 +38,6 @@ ExchangeParamType = Union["AbstractExchange", str]
 ConsumerTag = str
 
 MILLISECONDS = 1000
-ZERO_TIME = datetime(1970, 1, 1)
 
 
 class SSLOptions(TypedDict, total=False):
@@ -958,6 +957,5 @@ __all__ = (
     "TransactionState",
     "UnderlayChannel",
     "UnderlayConnection",
-    "ZERO_TIME",
     "get_exchange_name",
 )
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -38,7 +38,7 @@ ExchangeParamType = Union["AbstractExchange", str]
 ConsumerTag = str
 
 MILLISECONDS = 1000
-ZERO_TIME = datetime.utcfromtimestamp(0)
+ZERO_TIME = datetime(1970, 1, 1)
 
 
 class SSLOptions(TypedDict, total=False):
--- a/aio_pika/message.py
+++ b/aio_pika/message.py
@@ -1,4 +1,3 @@
-import time
 import warnings
 from datetime import datetime, timedelta, timezone
 from functools import singledispatch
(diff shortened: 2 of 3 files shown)


decaz:
Commit [e966a96](https://github.com/mosquito/aio-pika/pull/610/commits/e966a965d079129d224fe64c6b2bd92199c4accb) may be reverted in the future when https://github.com/gmr/pamqp/issues/51 will be released (if it will be merged).

Alviner (approved):

aio_pika/abc.py
```diff
@@ -38,7 +38,7 @@
 ConsumerTag = str
 
 MILLISECONDS = 1000
-ZERO_TIME = datetime.utcfromtimestamp(0)
+ZERO_TIME = datetime(1970, 1, 1)
```
mosquito, line 41: Might be better use timezone here too?

decaz: This constant is used only here: https://github.com/mosquito/aio-pika/blob/2e63f3a8266353757e3a922b90a1602225110818/aio_pika/message.py#L61-L63

So it should stay as naive as another operand.

@mosquito concerning this part of code: is `*t[:7]` right? I guess it should be `*t[:6]` as in other functions in this module..

decaz:
@mosquito https://github.com/gmr/pamqp/pull/52 was merged and released within version 3.3.0. I think firstly aiormq should be upgraded to pamqp of this new version and after that I can remove commit https://github.com/mosquito/aio-pika/pull/610/commits/e966a965d079129d224fe64c6b2bd92199c4accb so this PR could be mergeable.

aio_pika/abc.py
mosquito:
```python
import calendar

@decode_expiration.register(time.struct_time) 
 def decode_expiration_struct_time(t: time.struct_time) -> float: 
     return calendar.timegm(t)
```

we should try to check something like this.

decaz: @mosquito I tidied up these single-dispatch functions within separate commit.

Issue #358, opened 2020-10-28
How to exit QueryIterator?


alex-oleshkevich:
I know it is not the first time it is posted here but still, I could not find a working solution.
My problem is that my application runs in CLI mode and doesn't react on ^C property. Here is an example:
```python
import asyncio

import aio_pika


async def main():
    async def get_connection():
        return await aio_pika.connect_robust('amqp://guest:guest@localhost')

    pool = aio_pika.pool.Pool(get_connection)
    async with pool.acquire() as connection:
        async with connection:
            channel = await connection.channel()
            queue = await channel.declare_queue('my_queue_name', durable=True)
            async with queue.iterator() as stream:
                async for message in stream:
                    print(message)


if __name__ == '__main__':
    asyncio.run(main())
```

At this moment everything works well. Messages are received and printed.
Then, when I send CTRL+C (SIGINT) to the process, my program does not close.
The messages are no longer received but the program runs.

I have to press ^C again and only then the process gets killed with this exception trace:

<details>
  <summary>Click to see the trace</summary>
```
^C^CTraceback (most recent call last):
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/asyncio/base_events.py", line 566, in run_until_complete
    self.run_forever()
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/asyncio/base_events.py", line 534, in run_forever
    self._run_once()
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/asyncio/base_events.py", line 1735, in _run_once
    event_list = self._selector.select(timeout)
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/selectors.py", line 468, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "amqp_test.py", line 24, in <module>
    asyncio.run(main())
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/asyncio/runners.py", line 46, in run
    _cancel_all_tasks(loop)
  File "/home/alex/.pyenv/versions/3.7.5/lib/python3.7/asyncio/runners.py", line 62, in _cancel_all_tasks
[...]

PR #615, opened 2024-01-17
QueueIterator raises StopAsyncIteration when iterator/channel is closed.

Darsstar:
See #358

Currently QueueIterator never throws a StopAsyncIterator exception. Not when the channel is closed, and not after QueueIterator's close method is called.
Which implies that starting an `async for message in queue.iterator():` loop will keep running "forever" even if no new message will ever arrive. ("forever" because the asyncio task can be canceled, etc.)

This PR fixes that in a backwards compatible way. Some tests are refactored to rely on this implicitly, and new ones that explicitly test QueueIterator.__anext__() throws certain exceptions have been added as well.


aio_pika/connection.py
```diff
@@ -201,8 +204,7 @@ async def ready(self) -> None:
     def __del__(self) -> None:
         if (
             self.is_closed or
-            self.loop.is_closed() or
-            not hasattr(self, "connection")
```
Darsstar, line 205: If I overlooked some method how `self` could end up with an `connection` attribute I will undo this change.

Darsstar:
@mosquito ping

mosquito:
@Darsstar this request contains a lot of changes that could potentially breaks backward compatibility. I'm still thinking about how to test it so that I understand I need to make a separate major release, or make do with a minor one.

Darsstar:
> this request contains a lot of changes that could potentially breaks backward compatibility.

I assume you are refering to:
- new abstract methods
- some 'protected' properties changing type (bool ->> asyncio.Event)

9.4.0 already dropped Python 3.7. The changes taking advantage of 3.8, all contained in a single commit, were made in a way it should be backward compatible.

It probably should be a new major version due to the protected properties.

Although, I think those could be rewritten as `@property`s with setters which notify a [asyncio.Condition](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition) ...
Than I think it boils down to wether you are supporting people inheriting from the abstract base classes without inheriting from the concrete classes aio-pika provides.
If you do the major version should be bumped.

aio_pika/queue.py
```diff
@@ -513,22 +528,61 @@ async def __aexit__(
     async def __anext__(self) -> IncomingMessage:
         if not hasattr(self, "_consumer_tag"):
             await self.consume()
+
+        if self._closed.is_set():
+            raise StopAsyncIteration
+
+        message = asyncio.create_task(
```
mosquito, line 535: Creating task on each message is significantly slower. Task creation works fine when the coroutine do some long job or waiting network.

Here, as many as three tasks are created that must not only be queued for execution, but also canceled upon the occurrence of one of the events.

mosquito (changes requested):
Thank you for not giving up on this improvement, I'm sorry that I don't have much time to devote to it right now. But I'm open to discussion.

Darsstar:
Small update: I am currently under the impression this PR isn't at fault.

I branched from `upstream/master` and altered the reconnect_stairway test to start at 60 and go all the way up to 4096, instead of every power of 2 in the range `[64, 4096]`. [The tests fail on all Python versions](https://github.com/Darsstar/aio-pika/actions/runs/8283763580/job/22667723294), so it seems to be some sort of race condition based on how tasks end up being scheduled...?
The larger the `stair` value, the less likely it is to fail.

I don't have a strong lead, but I started looking at `aiomisc` and `aiormq` as well. Hopefully I'll stumble on the cause in the next two weeks or so...

aio_pika/queue.py
Darsstar: As of Python 3.11 passing coroutines to `asyncio.wait()` is forbidden. So I recuded it to one task per message, instead of three.

I don't think it is possible to remove that last task without effectively reimplementing `asyncio.wait()`.

Darsstar:
The tests are passing again :)

Public backward incompatible API changes:
- QueueIterator.close()` no longer accepts arguments

Internal backward incompatible API changes:
- `Connection._closed` changed from being of type `bool` to type `asyncio.Future`
- `Channel._closed` changed from being of type `bool` to type `asyncio.Future`

Those are all I found going over all the changes again.

My preference would be leaving it as is and reflecting that in the version number. Say the words and I will will turn `Connection._closed` and `Channel._closed` into properties and undo the `QueueIterator.close()` change.

LockedThread:
Hey @Darsstar, I am coming from https://github.com/mosquito/aio-pika/issues/623 and I have a question about how we should be handling the `TimeoutError` being raised from the application perspective? Recently, this exception has been throwing periodically in my production code and I am unsure of how I should be handling it... I thought the `RobustConnection` would handle this.

LockedThread:
I just re-read the issue and this PR may also be fixing a memory leak I have been facing for awhile. I am going to pull this version down and see if it fixes it. Will report back.

LockedThread:
This PR also seems to fix my memory leak. @mosquito Do we have an ETA on when this could be merged? I am going to start using his version now.

mosquito:
@LockedThread @Darsstar sorry, lots of changes, should planning to retest all these myself.

LockedThread:
> @LockedThread @Darsstar sorry, lots of changes, should planning to retest all these myself.

All good, we appreciate your diligent work in supporting this project.

Darsstar:
Unsurprisingly fixing this requires more code and therefor increases the runtime.

@gglluukk's [benchmark](https://github.com/mosquito/aio-pika/pull/627#issuecomment-2081255505) (thanks!) show about 10% runtime degredation:
```
before your patch:
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #1: 4.905
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #2: 4.503
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #3: 4.809
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #4: 4.870
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #5: 4.741
Average: 4.766

after your patch:
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #1: 5.481
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #2: 5.123
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #3: 5.397
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #4: 5.437
rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #5: 5.128
Average: 5.313
```

LockedThread:
> Unsurprisingly fixing this requires more code and therefor increases the runtime.
> 
> 
> 
> @gglluukk's [benchmark](https://github.com/mosquito/aio-pika/pull/627#issuecomment-2081255505) (thanks!) show about 10% runtime degredation:
> 
> ```
> 
> before your patch:
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #1: 4.905
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #2: 4.503
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #3: 4.809
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #4: 4.870
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #5: 4.741
> 
> Average: 4.766
> 
> 
> 
> after your patch:
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #1: 5.481
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #2: 5.123
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #3: 5.397
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #4: 5.437
> 
> rmq2psql.py --loop-type queue_iteration_with_timeout --max-bulks 100: #5: 5.128
> 
> Average: 5.313
> 
> ```

It would be great to get the time down but I am personally willing to accept a 10% degradation in order to not have a memory leak.

Darsstar:
> It would be great to get the time down but I am personally willing to accept a 10% degradation in order to not have a memory leak.

I don't see optimisation potential without porting the library to C/C++/Rust. (While keeping the implementation correct.)

LockedThread:
> > It would be great to get the time down but I am personally willing to accept a 10% degradation in order to not have a memory leak.
> 
> 
> 
> I don't see optimisation potential without porting the library to C/C++/Rust. (While keeping the implementation correct.)

Hopefully PyO3 gets better support for async/await, while maintaining interoperability with asyncio. There is a lot of active work on this right now. One that happens, using Rust will be feasible.

In the meantime, losing 10% in performance is worthwhile to make sure I dont get OOM kills.

mosquito:
You should do a performance test with cProfile for example. The last time I did this, it marshall in pamqp the slowest one.

aio_pika/queue.py
Darsstar: I was wrong.

No more task for message. In fact, I skipped the `await` if there are messages in the queue and use the `get_nowait` function instead.

mosquito:
I guess we should just finish to this already. Will you fix the code style?

Darsstar:
> I guess we should just finish to this already. Will you fix the code style?

Done

Darsstar:
I finally profiled it...

<details>
<summary>Python file I used to profile</summary>

```
import asyncio
import cProfile

from aio_pika import Message
from aio_pika.robust_connection import connect_robust

NUMBER_OF_MESSAGES = 1_000_000
# NUMBER_OF_MESSAGES = 1_000


async def main() -> None:
    profiler = cProfile.Profile(
        timer=asyncio.get_event_loop().time,
    )

    async with await connect_robust(host='localhost', port=5672, login='guest', password='guest') as conn:
        async with await conn.channel() as channel:
            exchange = channel.default_exchange
            queue = await channel.declare_queue('queue')

            pending = set()

            for i in range(NUMBER_OF_MESSAGES):
                pending.add(
                    asyncio.create_task(
                        exchange.publish(
                            message=Message(
                                body=f"message {i}".encode(encoding='utf-8'),
                            ),
                            routing_key=queue.name,
                            mandatory=True,
                        )
                    )
                )

            await asyncio.wait(pending)

            profiler.enable()

            count = 0

            async with queue.iterator() as queue_iter:
                async for message in queue_iter:
                    await message.ack()
                    count += 1
                    if count == NUMBER_OF_MESSAGES:
                        await queue_iter.close()
                        break

            profiler.disable()
            profiler.print_stats()
            profiler.dump_stats('D:/PR.stats')

if __name__ == '__main__':
    asyncio.run(main())
```

</details>

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
# before
  1000001    2.035    0.000   15.227    0.000 D:\Projects\aio-pika\aio_pika\queue.py:505(__anext__)
#after
  1000001    3.890    0.000    9.679    0.000 D:\Projects\aio-pika\aio_pika\queue.py:552(__anext__)
```

Regarding `tottime`: there is more code in `QueueIterator.__anext__` in `after`, so it is logical more cycles are spent in that frame.
Regarding `cumtime`: yay! A goog 5.5 seconds faster :)

I also made `CallbackCollection` inherit from `Generic` to track down a bug I introduced, and it led me to an existing bug.

decaz (approved):

walkingpendulum:
gentle ping with this PR, guys. any ETA on the merge? looking forward to see this released

Darsstar:
@mosquito think you will have time before the end of the year? (It's been a month since the last ping, and the end of the year is about a months away if I factor in holidays. No more significance than that.)

If a face to face review has your preference that seems doable. My work is one metro stop away from the Nebius offices. I don't mind paying for a drink.

LockedThread:
> @mosquito think you will have time before the end of the year? (It's been a month since the last ping, and the end of the year is about a months away if I factor in holidays. No more significance than that.)
> 
> 
> 
> If a face to face review has your preference that seems doable. My work is one metro stop away from the Nebius offices. I don't mind paying for a drink.

I am definitely willing to chip into the beer fund if that means this gets merged!!!

mosquito (approved):

mosquito:
Thanks guys, sorry for taking so long to respond. These changes have already been published in version [9.5.0](https://pypi.org/project/aio-pika/9.5.0/)

Issue #630, opened 2024-05-17
Messages are NACK'ed multiple times around queue deletion


lfse-slafleur:
Hello!

Thank you for aio-pika, it is a great library.

I have been running into some issues when using aio-pika in the orchestrator component for another opensource project called OMOTES (https://github.com/Project-OMOTES/orchestrator/). We use aio-pika to create quite a few queue's, receive 1 or more messages on the queue and delete the queue after. Now, we have found an issue where occassionally Rabbitmq (3.12.14-management) raises a `PRECONDITION_FAILED - unknown delivery tag`  exception to the subscriber of the queue's.

Using wireshark, I have managed to analyze the issue. Attached: [rabbitmq_unknown_delivery_tag.zip](https://github.com/mosquito/aio-pika/files/15351123/rabbitmq_unknown_delivery_tag.zip)

Context:
- There is 1 subscriber and 1 publisher. We are only interested in the subscriber which we can filter using `tcp.port == 49146`
- There is only 1 channel.
- Subscriber uses a prefetch of 1. 
- The subscriber creates a bunch of queue's in the channel with names `test_{0,1,2,3,4,5,6,7,8,9}`.
- Once the subscriber has created all queue's, it sends a message to the publisher that it is ready by sending `ready!` to the queue `ready`.
- After this, the publisher will send 5 messages to each of the `test_{0,1,2,3,4,6,7,8,9}` queues.
- The subscriber will delete the queue after receiving 1 message on each of the queues, also deleting the other 4 messages without being read/delivered.
- This forces a situation where there are more messages available in the queue for the subscriber even though it will only ACK 1 message. Also, the subscriber may have received more messages which it may NACK.

This leads to the issue as analyzed in the attached wireshark pcap:

- Frame 267 shows the `PRECONDITION_FAILED - unknown delivery tag 16` being raised by RabbitMQ.
- This is in relation to the NACK on frame 266 where `delivery tag 16` is NACK'ed.
- However, in the earlier frame 263 every delivery up to and including tag 17 was NACK'ed, which includes 16.
- This is the cause of the `unknown delivery tag` error as delivery tag 16 is now NACK'ed twice.

This is also documented here: https://www.rabbitmq.com/amqp-0-9-1-reference#basic.nack.multiple

Couple of more comments regarding this issue:
- Delivery tag 16 and 17 happen in frame 225.
[...]

PR #634, opened 2024-06-25
630: Only nack messages upon cancellation of a consumer subscription …

lfse-slafleur:
…that belong to that consumer subscription.

### DIFF
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -441,7 +441,7 @@ class QueueIterator(AbstractQueueIterator):
 
         log.debug("Queue iterator %r closed", self)
 
-        # Reject all messages
+        # Reject all messages that have been received and in the buffer/cache.
         while not self._queue.empty():
             msg = self._queue.get_nowait()
 
--- a/aio_pika/queue.py
+++ b/aio_pika/queue.py
@@ -442,13 +442,8 @@ class QueueIterator(AbstractQueueIterator):
         log.debug("Queue iterator %r closed", self)
 
         # Reject all messages
-        msg: Optional[IncomingMessage] = None
-        try:
-            while True:
-                msg = self._queue.get_nowait()
-        except asyncio.QueueEmpty:
-            if msg is None:
-                return
+        while not self._queue.empty():
+            msg = self._queue.get_nowait()
 
             if self._amqp_queue.channel.is_closed:
                 log.warning(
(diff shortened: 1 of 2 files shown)


lfse-slafleur:
Closes #630

tests/test_amqp.py
```diff
@@ -1587,6 +1587,141 @@ async def test_heartbeat_disabling(
         async with connection:
             assert heartbeat == 0
 
+    async def test_non_acked_messages_are_redelivered_to_queue(
```
lfse-slafleur, line 1590: Test to see if messages are properly nack'ed and I didn't break anything.

tests/test_amqp.py
```diff
@@ -1587,6 +1587,141 @@ async def test_heartbeat_disabling(
...
+
+            assert incoming_message.body == all_bodies[i]
+
+        # Check if the queue is now empty
+        assert await queue.get(fail=False, timeout=.5) is None
+
+        # Cleanup, delete the queue
+        await queue.delete()
+
+    async def test_regression_only_messages_cancelled_subscription_are_nacked(
```
lfse-slafleur, line 1636: Regression test which failed before fixing the code. Documentation is inline to show why the test is constructed the way it is.

aio_pika/queue.py
```diff
@@ -441,32 +441,24 @@ async def close(self, *_: Any) -> Any:
...
 
-        # Reject all messages
-        msg: Optional[IncomingMessage] = None
-        try:
-            while True:
-                msg = self._queue.get_nowait()
-        except asyncio.QueueEmpty:
-            if msg is None:
-                return
+        # Reject all messages that have been received and in the buffer/cache.
```
lfse-slafleur, line 444: While I fixed the code here locally as I thought it makes sense, something does still is uneasy with me. I would expect that the code here is not necessary and may even contain a liveliness issue.

Regarding the liveliness issue: Assume a queue that is receiving messages rapidly. The 'while' loop would be stuck manually nack'ing each and every single message until it is fast enough to escape the 'while' loop. It is possible this never happens. Also, it is possible that messages are still received after escaping the 'while' loop but just before the subscription termination message is send to RabbitMQ. These messages are not nack'ed.

Regarding the necessity: If an application wants each message to arrive, the message would not be automatically ack'ed but rather manually ack'ed after the message is persisted by whatever application integrates aio-pika. If a different delivery guarantee is expected, I would expect the application to allow for messages being dropped or duplicated. In other words, if the local storage is not explicitely ack'ed, the whole system should still work 'as expected'. ('As expected' here means as configured. If is a message is auto-acked (at-most-once guarantee), the application should be able to deal with missing messages. If at-least-once or exactly-once is expected, RabbitMQ should just requeue the moment the connection is dropped even if we do not requeue the messages).

So I am a bit confused as to why this code is required. @mosquito  Perhaps you could elaborate / provide guidance? Thanks in advance!

mosquito (approved):

mosquito:
@lfse-slafleur released as 9.4.2

PR #658, opened 2024-12-05
fix: RobustChannel should not reopen after close() call

Tasssadar:
It's not possible to close the RobustChannel right now, it always reopens.

### DIFF
--- a/aio_pika/robust_channel.py
+++ b/aio_pika/robust_channel.py
@@ -121,6 +121,14 @@ class RobustChannel(Channel, AbstractRobustChannel):    # type: ignore
 
         return exc
 
+    async def close(
+        self,
+        exc: Optional[aiormq.abc.ExceptionType] = None,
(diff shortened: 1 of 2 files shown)


przemyslaw-jakielaszek:
@mosquito could you please review it? 🙏 

In aio_pika>=9.5.0, a memory leak occurs when using robust channels due to the recovery of channels after closure. This issue is particularly detrimental for applications acting as publishers that open numerous channels.

Example code:
```
def publish(message):
    async with self._connection.channel() as channel:
        ...
```

Above code fails only because of comment line is too long.

Thank you for your help.

mosquito (approved):

mosquito:
@Tasssadar please fix linter issues and we will ready to merge this

przemyslaw-jakielaszek:
@mosquito I think something didn't work here as new release wasn't released to PYPI.

PR #661, opened 2024-12-26
readme: add type annotation to connection in consumer example

snosratiershad:


s-banach:
This isn't the fix. The problem is the annotation in aio-pika.

It should be annotated like this:


```python
T = TypeVar("T")

async def connect_robust(
    url: Union[str, URL, None] = None,
    *,
    ...
    connection_class: Type[T] = RobustConnection,
    **kwargs: Any,
) -> T: ...
```

mosquito:
@s-banach let discuss it in a another PR, please. I have to merge this

PR #665, opened 2025-02-17
fix AbstractQueueIterator does not match contextlib.AbstractAsyncContextManager

### LINKED ISSUE #664: AbstractQueueIterator does not match contextlib.AbstractAsyncContextManager
Class `AbstractQueueIterator`, seemingly has wrong `__aenter__` method's signature.

```python
    @abstractmethod
    def __aenter__(self) -> Awaitable["AbstractQueueIterator"]:
        raise NotImplementedError
```

For some reason, async is missing here. Although, it should be here according to contextlib context manager:

```python
class AbstractAsyncContextManager(abc.ABC):

    """An abstract base class for asynchronous context managers."""

    __class_getitem__ = classmethod(GenericAlias)

    async def __aenter__(self):
        """Return `self` upon entering the runtime context."""
        return self

    @abc.abstractmethod
    async def __aexit__(self, exc_type, exc_value, traceback):
        """Raise any exception triggered within the runtime context."""
        return None

    @classmethod
    def __subclasshook__(cls, C):
        if cls is AbstractAsyncContextManager:
            return _collections_abc._check_methods(C, "__aenter__",
                                                   "__aexit__")
        return NotImplemented
```

And is then implemented in `QueueIterator`,

```python
class QueueIterator(AbstractQueueIterator):
    DEFAULT_CLOSE_TIMEOUT = 5
    # ...
    async def __aenter__(self) -> "AbstractQueueIterator":
        if not hasattr(self, "_consumer_tag"):
            await self.consume()
        return self
```

Not big of an issue since abstract classes are never instantiated, but still.

Python 3.9, aio-pika==9.5.4

Quakumei:
Resolves #664

### DIFF
--- a/aio_pika/abc.py
+++ b/aio_pika/abc.py
@@ -387,7 +387,7 @@ class AbstractQueueIterator(AsyncIterable[AbstractIncomingMessage]):
         raise NotImplementedError
 
     @abstractmethod
-    def __aenter__(self) -> Awaitable["AbstractQueueIterator"]:
+    async def __aenter__(self) -> Awaitable["AbstractQueueIterator"]:
         raise NotImplementedError
 
     @abstractmethod


aio_pika/abc.py
```diff
@@ -387,7 +387,7 @@ def __aiter__(self) -> "AbstractQueueIterator":
         raise NotImplementedError
 
     @abstractmethod
-    def __aenter__(self) -> Awaitable["AbstractQueueIterator"]:
+    async def __aenter__(self) -> Awaitable["AbstractQueueIterator"]:
```
Darsstar, line 390: now the `Awaitable` should be removed

PR #673, opened 2025-05-12
Add global_ param to basic_qos call in RobustChannel _on_open method

### LINKED ISSUE #583: The global_ param to basic_qos is not set on RobustChannel reopen
In the call to `aiormq.Channel.basic_qos` inside `RobustChannel._on_open`, the values that were stored in `self._prefetch_count` and `self._prefetch_size` from the call to `set_qos` are passed, but not `self._global_qos`:

```Python
await channel.basic_qos(
    prefetch_count=self._prefetch_count,
    prefetch_size=self._prefetch_size,
)
```

Is this intentional? I don't know why you wouldn't want to pass `global_` as well, like:

```Python
await channel.basic_qos(
    prefetch_count=self._prefetch_count,
    prefetch_size=self._prefetch_size,
    global_=self._global_qos,
)
```

jeff-foster-trabon:
global_ was sent in set_qos, and cached in self._global_qos, but was not sent in subsequent qos calls upon reconnect, leading to potentially different channel behavior after reconnect

Fixes #583

### DIFF
--- a/aio_pika/robust_channel.py
+++ b/aio_pika/robust_channel.py
@@ -143,6 +143,7 @@ class RobustChannel(Channel, AbstractRobustChannel):    # type: ignore
         await channel.basic_qos(
             prefetch_count=self._prefetch_count,
             prefetch_size=self._prefetch_size,
+            global_=self._global_qos,
         )
 
         for exchange in exchanges:


decaz (approved):
Looks good to me.

PR #680, opened 2025-06-27
add robust_connection_multihost.py

federicsp:
Add RobustConnectionMultiHost class for fault-tolerant RabbitMQ connections:

- Implements a connection class that tries multiple broker URLs in sequence until one succeeds.

- Automatically handles reconnection on unexpected connection closures.

### DIFF
--- a/aio_pika/robust_connection_rrhost.py
+++ b/aio_pika/robust_connection_rrhost.py
@@ -16,16 +16,18 @@ class RobustConnectionRRHost:
     cycling through provided URLs until a successful connection is made.
     """
 
-    def __init__(self, urls: List[str], default_port: int = 5672,
-                 **kwargs: Any):
+    def __init__(
+        self, urls: List[str], default_port: int = 5672, **kwargs: Any
+    ):
         """
-        Initialize with a list of broker URLs, normalizing and applying default port if missing.
+        Initialize with a list of broker URLs, normalizing and applying
+        default port if missing.
 
         :param urls: List of AMQP broker URLs as strings.
         :param default_port: Default port used if not specified in URLs.
         :param kwargs: Additional arguments passed to RobustConnection.
         """
-        self.urls = []
+        self.urls: List[URL] = []
         for url in urls:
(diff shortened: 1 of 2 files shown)


aio_pika/robust_connection_multihost.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __init__(self, urls: List[str], default_port: int = 5672, **kwargs: Any):
+        self.urls = [self._prepare_url(url, default_port) for url in urls]
+        self.kwargs = kwargs
+        self._current_url: Optional[URL] = None
+        self._connect_timeout: Optional[float] = None
+        super().__init__(str(self.urls[0]), **kwargs)
+
+    def _prepare_url(self, url: str, default_port: int) -> URL:
+        if not url.startswith("amqp://"):
```
mosquito, line 36: what about `amqps`?

aio_pika/robust_connection_multihost.py
```diff
@@ -0,0 +1,81 @@
+from typing import List, Optional, Any
+from yarl import URL
+from aio_pika.robust_connection import RobustConnection
+import asyncio
+
+class RobustConnectionMultiHost(RobustConnection):
```
mosquito, line 6: `RobustConnectionRRHost` semantically it is closer, because you connection is not a multihost, it is a round-robin retying to connect only to one.

aio_pika/robust_connection_multihost.py
```diff
@@ -0,0 +1,81 @@
...
+                last_exc = e
+        raise last_exc or RuntimeError("All connection attempts failed")
+
+    async def _on_connection_close(self, closing):
+        if not self.is_closed:
+            await self.reconnect()
+        await super()._on_connection_close(closing)
+
+    async def reconnect(self):
+        await asyncio.sleep(2)
```
mosquito, line 75: Why `2`? Why not `60` or `0.0001`?

mosquito (changes requested):

PR #699, opened 2026-03-28
fix: use TypeVar for connect and connect_robust return types

### LINKED ISSUE #683: Incorrect type annotation for connect_robust return value
# Incorrect type annotation for connect_robust return value

## Bug Description

The return type annotation for `aio_pika.connect_robust` is currently `AbstractRobustConnection`, but in practice, when passing `connection_class=aio_pika.RobustConnection`, the returned object is an instance of `RobustConnection`. This causes a type mismatch error in static type checkers like mypy when annotating the result as `aio_pika.RobustConnection`, even though the runtime type is correct. The type signature of `connect_robust` does not reflect the actual return type when a specific `connection_class` is provided, leading to false-positive type errors.

## How to Reproduce

I prepared a minimal reproducible example to demonstrate the issue.

```python
# /// script
# requires-python = ">=3.10"
# dependencies = ["aio-pika"]
# ///
import asyncio

import aio_pika


async def loop() -> None:
    connection: aio_pika.RobustConnection = await aio_pika.connect_robust(
        "amqp://guest:guest@127.0.0.1/",
        connection_class=aio_pika.RobustConnection,
    )
    print(type(connection))  # <class 'aio_pika.robust_connection.RobustConnection'>


def main() -> None:
    asyncio.run(loop())


if __name__ == "__main__":
    main()

```

```sh
> uvx mypy main.py
main.py:11: error: Incompatible types in assignment (expression has type "AbstractRobustConnection", variable has type "RobustConnection")  [assignment]
        connection: aio_pika.RobustConnection = await aio_pika.connect_robust(
                                                ^
Found 1 error in 1 file (checked 1 source file)
```

## Expected Behavior

I expect the `connection` variable to have `aio_pika.RobustConnection` type.

hasansezertasan:
## Summary

- Use `TypeVar` bound to `AbstractConnection` / `AbstractRobustConnection` for the `connection_class` parameter and return type of `connect()` and `connect_robust()`
- This allows mypy to correctly infer the return type when a concrete `connection_class` is provided (e.g., `aio_pika.RobustConnection`)

Closes #683

## Test plan

- [ ] Verify `mypy` no longer reports type mismatch when annotating the result of `connect_robust(connection_class=RobustConnection)` as `RobustConnection`
- [ ] Verify default calls (without `connection_class`) still type-check correctly
- [ ] Run existing test suite to ensure no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/aio_pika/connection.py
+++ b/aio_pika/connection.py
@@ -12,6 +12,7 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    overload,
 )
 
 
@@ -36,6 +37,7 @@ from .tools import CallbackCollection
 
 log = get_logger(__name__)
 T = TypeVar("T")
+ConnectionType = TypeVar("ConnectionType", bound=AbstractConnection)
 
 
 class Connection(AbstractConnection):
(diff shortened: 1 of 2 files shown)


aio_pika/connection.py
decaz, line 358: Couldn't `ConnectionType` be used here instead of these overloads?

Same question is for `RobustConnectionType` case.

decaz (commented):
@hasansezertasan thanks for PR! Take a look at my comment please.

aio_pika/connection.py
hasansezertasan: The overloads are needed here. Without them, calling `connect()` without a `connection_class` argument would make mypy resolve `ConnectionType` to its upper bound (`AbstractConnection`) rather than the concrete `Connection` type — since mypy can't infer a TypeVar from a default parameter value.

The two overloads let us express:
1. No `connection_class` → returns `Connection` (specific concrete type)
2. `connection_class: Type[ConnectionType]` provided → returns `ConnectionType` (generic, inferred from the argument)

If we used just `connection_class: Type[ConnectionType] = Connection` in a single signature, users would need to annotate the result manually when not passing `connection_class`, which defeats the purpose of the fix.

Same reasoning applies to the `RobustConnectionType` / `connect_robust` case.

mosquito (approved):

decaz:
@hasansezertasan thank you!

PR #703, opened 2026-07-09
Fix overloads of `connect` and `connect_robust` methods

decaz:
Overloads are missing parameter `kwargs` of methods.


decaz:
Support for `default` value of `TypeVar` was added in 3.13 so it's impossible to merge current implementation, - we need another solution.

@hasansezertasan FYI

